PR #24, opened 2014-11-06
Fixes embedded dylibs/frameworks only run on iOS8 warning by ignoring linker warnings

artman:
XCode6.1 (falsely) warns that dylibs don't work on <iOS8. Suppressing this warning with linker flags for now.

### DIFF
--- a/Starscream.xcodeproj/project.pbxproj
+++ b/Starscream.xcodeproj/project.pbxproj
@@ -429,6 +429,10 @@
 				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
 				IPHONEOS_DEPLOYMENT_TARGET = 7.0;
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks";
+				OTHER_LDFLAGS = (
+					"-Xlinker",
+					"-w",
+				);
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				PROVISIONING_PROFILE = "";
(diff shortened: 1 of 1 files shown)


pizthewiz:
While dynamic frameworks may run on iOS 7, App Store submission will be _rejected_ (TestFlight too) if a bundled framework is not built with a deployment target of iOS 8+, see realm/realm-cocoa#900 and Hearst-DD/ObjectMapper#23 for other cases of this. The realm team worked around this by creating a static framework target for iOS 7 and dynamic framework target for iOS 8 use, or one could of course just pull the files in directly.

So quieting the warning masks the issue a bit no?

daltoniam:
@pizthewiz that is a good point. Personally for apps that I have to support iOS 7 on, I have just been pulling the files in directly. I know that isn't ideal, but since iOS 7 is going probably fall off as a "supported development version" when iOS 9 drops, it might be an ok compromise. If there is the demand (or a pull request), I could see creating a static lib like realm has done.

pizthewiz:
> for apps that I have to support iOS 7 on, I have just been pulling the files in directly

Yep, that technique has been suggested for the ObjectMapper library too. It could be more formalized by setting the framework deployment target to iOS 8 and add a mention in the README - this would at least put iOS 8 on the fully supported path. The current form of the framework target is not usable for App Store submission for iOS 7 _or_ iOS 8, and suppressing the warnings just further obscures the core issue no?

daltoniam:
Agreed. The README should probably update to reflect this. That is a good point, I will change the framework to allow this as well. Thanks for bring it up.

Issue #42, opened 2014-12-19
WebSocketDelegate's methods should receive a websocket object as it's first parameter


PrideChung:
Something like this:

``` swift
public protocol WebSocketDelegate: class {
    func webSocketDidConnect(webSocket: WebSocket)
    func webSocket(webSocket: WebSocket, didDisconnect error: NSError?)
    func webSocket(webSocket: WebSocket, didWriteError error: NSError?)
    func webSocket(webSocket: WebSocket, didReceiveMessage text: String)
    func webSocket(webSocket: WebSocket, didReceiveData data: NSData)
}
```

The delegate object may has no reference to the `WebSocket` and it may need the `webSocket` parameter to do it's work. Or in another case, the delegate object is a delegate of multiple `WebSocket`s, the `webSocket` parameter can be used to distinguish which `WebSocket` is triggering this event.

PR #47, opened 2015-01-25
reference to calling websocket instance

cyphercommunications:
Not sure if this would detract from the clean interface you have, but I've found it to be useful

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ After you are connected, there are some delegate methods that we need to impleme
 websocketDidConnect is called as soon as the client connects to the server.
 
 ```swift
-func websocketDidConnect() {
+func websocketDidConnect(socket: WebSocket) {
     println("websocket is connected")
 }
 ```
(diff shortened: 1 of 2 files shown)


acmacalister:
@cyphercommunications appreciate the PR. What where some of the use cases you found useful exposing the underlining socket in the interface?

cyphercommunications:
We use a hybrid app on iOS that requires multiple voip web sockets to be used. We have written a wrapper in javascript that allows the web view to use the existing browser WebSocket api, but still keep the socket listening while the device sleeps.

It may not be the best technique, but we found exposing the underlining socket to be necessary in order to know which socket to trigger the correct handlers for on the javascript side.

nicolasbigot:
+1 for this, this is something I really need too (one unique delegate for multiple WebSocket instances)

daltoniam:
Alright, I concede :smile: . Just for reference this was discussed in #42 as well.

PR #50, opened 2015-02-05
allow changing the queue for executing delegate callbacks

wentao:


daltoniam:
Looks good. Only thing I notices is need to add ```public``` keyword to the ```queue``` variable so it has the proper visibility in the framework. After that I think we are good to merge.

wentao:
Yeah, public modifier is added to the queue var.

Issue #55, opened 2015-02-12
Crash EXC_BREAKPOINT code = 1 line 378


vani2:
I noticed a strange bug. Now I have no idea to fix it.

```
* thread #30: tid = 0xd7890, 0x00000001000caf90 WeLikeChat`WeLikeChat.WebSocket.(buffer=(value = Builtin.RawPointer = 0x00000001600e2351), bufferLen=Swift.Int at 0x0000000100cf56a8, self=0x000000015fe89b20) (WeLikeChat.WebSocket)(Swift.UnsafePointer<Swift.UInt8>, bufferLen : Swift.Int) -> () + 17052 at WebSocket.swift:378, queue = 'com.apple.root.default-qos', stop reason = EXC_BREAKPOINT (code=1, subcode=0x1000caf90)
  * frame #0: 0x00000001000caf90 WeLikeChat`WeLikeChat.WebSocket.(buffer=(value = Builtin.RawPointer = 0x00000001600e2351), bufferLen=Swift.Int at 0x0000000100cf56a8, self=0x000000015fe89b20) (WeLikeChat.WebSocket)(Swift.UnsafePointer<Swift.UInt8>, bufferLen : Swift.Int) -> () + 17052 at WebSocket.swift:378
    frame #1: 0x00000001000cb44c WeLikeChat`WeLikeChat.WebSocket.(buffer=(value = Builtin.RawPointer = 0x00000001600e2351), bufferLen=Swift.Int at 0x0000000100cf5800, self=0x000000015fe89b20) (WeLikeChat.WebSocket)(Swift.UnsafePointer<Swift.UInt8>, bufferLen : Swift.Int) -> () + 320 at WebSocket.swift:566
    frame #2: 0x00000001000caf60 WeLikeChat`WeLikeChat.WebSocket.(buffer=(value = Builtin.RawPointer = 0x00000001600e2243), bufferLen=Swift.Int at 0x0000000100cf6cc8, self=0x000000015fe89b20) (WeLikeChat.WebSocket)(Swift.UnsafePointer<Swift.UInt8>, bufferLen : Swift.Int) -> () + 17004 at WebSocket.swift:555
[...]

PR #58, opened 2015-02-16
Add surfacing of code sent when server closes connection with specific code

robskillington:
- Call delegate didDisconnect when server closes connection with code
- Avoid crashing when outputStream is nil when server closes with code

### DIFF
--- a/WebSocket.swift
+++ b/WebSocket.swift
@@ -42,6 +42,11 @@ public class WebSocket : NSObject, NSStreamDelegate {
         case MessageTooBig          = 1009
     }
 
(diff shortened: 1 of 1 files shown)


robskillington:
Hey @daltoniam I had to make these changes to be able to read the code that the server was sending when closing the connection.  Also it was crashing as the outputStream would be nil and it would try to `writeError` so I added protection to that code path.

daltoniam:
@robskillington you rock, thanks! I appreciate you taking the time to track this down. This might also fix #55.

PR #62, opened 2015-03-02
Update syntax to Swift 1.2

mogstad:
Fixed the last errors caused by the altered Syntax in Swift 1.2


daltoniam:
Awesome, thank you!

PR #63, opened 2015-03-04
Yodeling Snowcat - Write ping frames

thebehera:
Allow snowcat to fire an astroid to the planet.
Allow client to write a ping frame to the server.

Here's a video for reference
http://youtu.be/Eu5ZJELRiJ8?t=42s

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,14 @@ The writeString method is the same as writeData, but sends text/string.
 self.socket.writeString("Hi Server!") //example on how to write text over the socket!
 ```
 
+### writePing
+
+The writePing method is the same as writeData, but sends a ping control frame.
+
+```swift
+self.socket.writePing(NSData()) //example on how to write a ping control frame over the socket!
+```
+
 ### disconnect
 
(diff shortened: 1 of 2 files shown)


daltoniam:
haha nicely done :smile:. Quick question, what is the use case of sending pings from the client? From my understanding normally the WebSocket server sends the pings and the client send a pong response. Also could you add a section in the README documenting the new method like the other public methods?

thebehera:
So in our use case the server needs to know who is in the background and who is in the foreground to filter our socket messages. We are using a voip background websocket and we don't want unnecessary data sent to the clients while in the background.

Also just for reference, in the [RFC 6455 Section 5.5.2](https://tools.ietf.org/html/rfc6455#section-5.5.2) it just specify "endpoint" rather than client or server. We've seen in other client libraries such as socket.io that clients ping and servers pong

@shanksauce is our backend guy and helped me write this explanation :+1: 

Committing an update soon to reflect the changes in the readme.

daltoniam:
Awesome, RFC reference FTW :smile: . Once the README is updated I will merge.

PR #67, opened 2015-04-07
fixed crash when inputStream and outputStream need to be closed

ndrppnc:
When you're in Airplane mode inputStream and outputStream can be nil and trying to force unwrapping them and then close them throws an exception and crashes the app. I wrapped them with an if block where I check if they're not nil first.


daltoniam:
Thanks!

PR #68, opened 2015-04-08
Check target correctly

mikajauhonen:
Would not build OS X version since UIKit is always imported. (Xcode 6.3 Beta 4)


daltoniam:
Ah, nice catch. Thanks!

PR #71, opened 2015-04-18
Fix .gitignore so that it ignores Carthage build directory as well

insanehunter:
This patch fixes a subtle issue with Carthage when Starscream is added as a submodule. Carthage creates a build directory, which makes submodule appear as dirty.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@
 
 # Xcode
 .DS_Store
-build/
+build
 *.pbxuser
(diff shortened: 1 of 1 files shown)


insanehunter:
This patch fixes a subtle issue with Carthage when Starscream is added as a submodule. Carthage creates a build directory, which makes submodule appear as dirty.

daltoniam:
:+1: thanks!

PR #73, opened 2015-04-21
[FIX] memory leak with the takeUnretainedValue()

yannickl:
Hi,

I just found a memory leak inside the `validateResponse` method with `takeUnretainedValue()` calls via the use of https://github.com/socketio/socket.io-client-swift.

Here the `CFHTTPMessageCreateEmpty` and the `CFHTTPMessageCopyAllHeaderFields` needs to be manage by swift so we need to call takeRetainedValue() instead.

Thank you for the great lib,

Best


daltoniam:
Awesome thanks! I was just thinking about this the other day that I should take a second look at that :smile:. Nice work :+1:

PR #79, opened 2015-05-10
Fix for documented force unwrapping of a potentially nil object

AndrewZurn:


daltoniam:
Nice catch :+1:. Thanks!

PR #88, opened 2015-05-28
Pong received delegate

thbaja:
I implemented a websocketDidReceivePongDelegate in a separate protocol.
This way it will be optional and will not break existing code where users would otherwise have to implement this method.

I am aware that it's possible to use `@objc` and `optional` but it comes with caveats. 
According to Ash Furrow it seems that separate protocols is the better solution: http://ashfurrow.com/blog/protocols-and-swift/

I'm only a few days into learning Swift so bear with me :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ func websocketDidReceiveData(socket: WebSocket, data: NSData) {
 }
 ```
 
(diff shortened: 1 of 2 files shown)


daltoniam:
Excellent work! I was going to suggest we implement it the same way as you suggest and you did it perfectly :smile: Even updated the README, much appreciated.

Issue #100, opened 2015-07-22
Swift 2.0 Errors


Stefandasbach:
With Swift 2.0 in Xcode Beta 4, there are several errors in the security file. Namely they stem from calling takeRetainedValue on certificates. It would be great if you could provide support for this.

daltoniam:
#101 should have fix this, let me know if that resolves it for you.

Stefandasbach:
Perfect, all is resolved.

PR #101, opened 2015-07-22
Xcode 7 / iOS 9 beta 4 compatibility

col:
Modified the Security class to make it compatible with Xcode 7 / iOS 9 beta 4.


daltoniam:
Awesome, thank you much. Should fix #100.

PR #104, opened 2015-08-24
expose disconnectStream to public for some cases we need to close and

lvvvvv:
cleanup the underlay stream when webscoket haven’t connected to remote server


daltoniam:
I don't understand the use case. Is there a reason you would call the underling `disconnectStream` method versus calling the `disconnect` method?

daltoniam:
@lvvvvv can you provide anymore detail on this?

PR #106, opened 2015-08-25
Update WebSocket.swift

Ewg777:
Xcode 7 beta 6 compatibility


daltoniam:
thanks! I already merged #108 which also implemented this same fix (just going through the emails in order :smile: ).

PR #110, opened 2015-08-26
Property for enabling only certain cipher suites for the WSS connection

lbanders:
We have the option of enabling only certain cipher suites to improve the security on the WSS connections. We have added a **enabledSSLCipherSuites** property which is an array of **SSLCipherSuite**. To disable, simply set to **nil**.

The selected cipher suites must be valid for the platform. If an error occurs while enabling the cipher suites for a stream, we will fail the connection.


daltoniam:
Cool, thanks for the PR. Three things: 

- can we remove the `NSLog` statements?
- can we just call `disconnectStream(error)` and not `doDisconnect(error)`? (I know it is like that on the other security stuff now ( [line 262](https://github.com/daltoniam/Starscream/blob/master/WebSocket.swift#L262) ), but I think that is unnecessary and can be removed.)
-   Can we return on both the failure cases after `disconnectStream`? We probably don't want it to start the runloop part when it wouldn't connect properly anyway.

Thanks! Any question please let me know.

lbanders:
Yes sure - I will updated the PR with those changes. They sounds like reasonable changes. 
We will probably also make one for setting the TLS protocols, but a PR will come for that later.

pcoltau:
I have corrected the code with respect to the review comments. Please review it again.

daltoniam:
looks good. Thanks!

PR #131, opened 2015-10-30
fixed connect() function using Master code

SamDecrock:
I don't know if you want this pull request.
I'm still using Swift 1.2, maybe there are others out there that might want the same


daltoniam:
good idea, probably a few folks out there. Thanks!

PR #133, opened 2015-11-13
clean up WebSocket.disconnect parameter

mayoff:
Now the parameter is properly an Optional<NSTimeInterval>, nil means no timeout, and zero means close immediately and forcibly.


daltoniam:
Thanks. Nice touch on the comments/docs, appreciate it.

PR #135, opened 2015-11-16
Sets UIRequiredDeviceCapabilities in tvOS target

simonbs:
arm64 must be set in UIRequiredDeviceCapabilities for tvOS targets when uploading to iTunes Connect.


daltoniam:
Good to know. Thanks!

PR #138, opened 2015-11-22
Make CloseCode enum public

Abizern:
Since they get passed to the websocketDidDisconnect delegate method,
making them public means that the codes can be switched against without
needing magic numbers.


daltoniam:
Ah good point, much cleaner. Thanks!

PR #140, opened 2015-11-24
Unwrap header safely when validating response.

markus-k:
When the `Sec-WebSocket-Accept` header is missing in the response, the app crashes due to an unwrapped nil value. With this pull request, the header is safely unwrapped and the app will not crash but simply disconnect.

The header can be missing when some proxy between the client and server eats this header (for whatever reason).


daltoniam:
Good point and nice catch. Thanks for the PR, I appreciate it.

PR #144, opened 2015-12-11
call websocketDidConnect delegate after connected is set to prevent race condition

JamesPrudente:
Attempting to write to the socket during websocketDidConnect fails silently sometimes because connected is not set until after processHTTP returns.  The isConnected guard in dequeueWrite can execute before connected is set to true.

This change ensures that connected is set before calling the delegate.

### DIFF
--- a/WebSocket.swift
+++ b/WebSocket.swift
@@ -348,7 +348,13 @@ public class WebSocket : NSObject, NSStreamDelegate {
         
         if !connected {
             connected = processHTTP(buffer, bufferLen: length)
-            if !connected {
+            if connected {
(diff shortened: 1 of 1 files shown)


daltoniam:
I think this will fix #147, correct?

JamesPrudente:
my PR is similar, but does not address the case in #147.  My PR is about websocketDidConnect being called after connected is set.  looks like the solution to #147 might be similar, but there are a bunch more cases where doDisconnect is called, so looks more complicated.

daltoniam:
Ok cool. Thanks!

Issue #123, opened 2015-09-28
Error: Invalid HTTP upgrade


wakinchan:
Hi,

I have random connection problem which display these logs:

```
websocket is connected
websocket is disconnected: "Invalid HTTP upgrade"
```

code:
```swift
    let useSSL = true
    let websocketScheme = "wss"
    let host = "x.x.x.x"
    let port: NSNumber = 8080
    
    public init() {
        let components = NSURLComponents()
        components.scheme = websocketScheme
        components.host = host
        components.port = port
        components.path = "/"
        
        webSocket = WebSocket(url: components.URL!)
        webSocket?.selfSignedSSL = useSSL
        webSocket?.delegate = self
        webSocket?.connect()
    }
```

It happens on iOS 9.0.1 (iPads).
With swift 1.x and Xcode 6.x everything worked well, but now it randomly failed.
Please give me some suggestion.
Thank you.

acmacalister:
Invalid Upgrade is just the response the server gives back when it couldn't switch the http connection over to WebSocket protocol. Do you have any way I can reproduce this? What server are you using? How often does it "randomly" happen? This doesn't give me much to go off or how to troubleshoot it, so I'm not sure what suggestions I can provide.

acmacalister:
We reran the the autobahn suite to ensure compliance. Good news is everything passed. We also tested a Ruby WebSocket server and can't reproduce this issue.

nuclearace:
I use this library in my [socket.io-client for swift](https://github.com/socketio/socket.io-client-swift) and I've had users complain that it's happening. What's odd is I've gotten complaints that it's happened on a Java based socket.io server and the default node.js server. My first thought was that something behind-the-scenes changed with NSInputStream/NSOutputStream that's causing these random failures. Personally I haven't been able to reproduce on OS X and everyone who's complained is using it on iOS.

nuclearace:
Also, I think this is where the actual message is coming from https://github.com/daltoniam/Starscream/blob/master/WebSocket.swift#L324 with the last things changing there being https://github.com/daltoniam/Starscream/blame/master/WebSocket.swift#L322-L324

acmacalister:
Thanks @nuclearace I read through the issue on your library. We tried the autobahn (Python) and Ruby based servers on iOS and couldn't reproduce the error either. Agreed. We are reviewing that code now.

daltoniam:
[...]

PR #145, opened 2015-12-11
Finds the HTTP Packet in the TCP stream even with empty content

ricardopereira:
#123

### DIFF
--- a/WebSocket.swift
+++ b/WebSocket.swift
@@ -388,10 +388,11 @@ public class WebSocket : NSObject, NSStreamDelegate {
         var k = 0
         var totalSize = 0
         for i in 0..<bufferLen {
+            totalSize = i
             if buffer[i] == CRLFBytes[k] {
                 k++
                 if k == 3 {
-                    totalSize = i + 1
+                    totalSize += 1
                     break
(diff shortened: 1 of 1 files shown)


ricardopereira:
Where can I write a test about this fix?

mattheworiordan:
Why not make that method a public method and inject a buffer example that passes and one that fails, but should pass, that this commit fixes?

ricardopereira:
@daltoniam I noticed that there are no tests implemented. What was the last time you run Autobahn Test? I run it yesterday and the result is: https://dl.dropboxusercontent.com/u/11377305/tmp/starscream-report-ws-20151213/clients/index.html. Some of them are failing.

daltoniam:
I haven't added any tests yet. My hope is to create a Swift HTTP/Websocket server that can be used for the tests, but that is going to take a fair amount of work, so I haven't gotten around to it yet. I will take a look at the autobahn tests.

daltoniam:
I was able to look into the failing tests the new `1.1.0` release should resolve this (hopefully we can get a test server created in Swift in the future to have better tests).

Issue #77, opened 2015-05-05
Starscream version 0.9.3 crashes on iPhone 5 and under


asafhlp:
```
Exception Type:  EXC_BAD_ACCESS (SIGBUS)
Exception Subtype: EXC_ARM_DA_ALIGN at 0x15b58c0a
Triggered by Thread:  2
Thread 2 name:  Dispatch queue: com.apple.root.default-qos
Thread 2 Crashed:
0   Starscream                    	0x004c9490 function signature specialization <Arg[0] = Exploded, Arg[1] = Exploded, Arg[2] = Owned To Guaranteed> of Starscream.WebSocket.(processRawMessage in _E91840E7069A0D2D302D042308E9D928) (Starscream.WebSocket)(Swift.UnsafePointer<Swift.UInt8>, bufferLen : Swift.Int) -> () (WebSocket.swift:493)
1   Starscream                    	0x004c6d1c function signature specialization <Arg[0] = Owned To Guaranteed> of Starscream.WebSocket.(processInputStream in _E91840E7069A0D2D302D042308E9D928) (Starscream.WebSocket)() -> () (WebSocket.swift:338)
2   Starscream                    	0x004c44b4 @objc Starscream.WebSocket.stream (Starscream.WebSocket)(ObjectiveC.NSStream, handleEvent : C.NSStreamEvent) -> () (WebSocket.swift:264)
3   CoreFoundation                	0x252de6d4 _signalEventSync + 144
```
Version 0.9.1 used to work with the exact use case.
Possibly related to the change from using 'byteSwapped' to using 'bigEndian' on UInt64.

Can I help further?

asafhlp:
Further inspection reveals this happens consistently on large packets (e.g. 190k~ characters). We haven't witnessed any issue on small packets.

daltoniam:
This sounds like #52. I tested on an iPad 2 with the largest messages possible and it worked properly. What kind of server are you using?

asafhlp:
We use Jetty for the WebSocket server. The issue doesn't happen on all devices, we verified it on the iPod 5 and iPhone 4S/5.  
I've looked into it more today. The crash happens only in release configuration, and not in debug configuration. After playing around some, it seems that setting
```SWIFT_OPTIMIZATION_LEVEL = "-Onone"```
for XCBuildConfiguration section under the Starscream.xcodeproj/project.pbxproj file relieves the issue.

daltoniam:
[...]

PR #146, opened 2015-12-11
Avoid doing unaligned reads/writes via UnsafePointer

mstorsjo:
Instead read/write individual bytes via UnsafePointer<UInt8>.

This fixes crashes that only happen when optimization is turned on,
on 32 bit arm builds.

Not all processors support unaligned reads/writes of 64 bit values.
Without optimizations, the reads/writes probably end up handling one
byte at a time (like this commit makes it do), while higher
optimization levels convert it to single reads/writes of the full
size, which may require that the pointer is properly aligned.

### DIFF
--- a/WebSocket.swift
+++ b/WebSocket.swift
@@ -434,6 +434,33 @@ public class WebSocket : NSObject, NSStreamDelegate {
         return false
     }
     
+    ///read a 16 bit big endian value from a buffer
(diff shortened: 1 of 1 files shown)


nuclearace:
Should fix #77

daltoniam:
Awesome, thank you much for fixing this. @nuclearace agreed.

PR #152, opened 2015-12-25
Fix #151

### LINKED ISSUE #151: Origin header should be optional and settable
According to the [RFC](https://tools.ietf.org/html/rfc6455):

> The request MUST include a header field with the name |Origin|
> [RFC6454] if the request is coming from a browser client.  If
> the connection is from a non-browser client, the request MAY 
> include this header field if the semantics of that client match
> the use-case described here for browser clients.  The value of
> this header field is the ASCII serialization of origin of the
> context in which the code establishing the connection is running.

steven851007:


daltoniam:
Looks good, nice work. Thank you much!

PR #154, opened 2016-01-06
fix for uncompleted headers in the buffer

dsato80:
processHTTP returns "false" when buffer contains uncompleted headers.
It needs to read further stream when two new lines are not found in the buffer.

I have encountered this situation with slow network connection (bluetooth tethering).
It may be related to #123


daltoniam:
Thanks for working on this. I believe we addressed this the `1.1.0` release. Let us know if you have any issues. Thanks again.

dsato80:
Thank you very much for working on this issue!
I found one missing line about fragBuffer. I will make another pull request.

PR #157, opened 2016-01-18
set fragBuffer data if processHTTP returns -1

dsato80:


daltoniam:
Good catch, nicely done. Thank you.

PR #160, opened 2016-01-22
Simple update to examples for Swift 2 (println() to print())

kylebshr:
I know this is small, but it seemed like a good idea to update the examples in the README to be valid in Swift 2 and greater.


daltoniam:
Agreed, can't believe I overlooked that. Small is still progress :smile:. Thanks!

kylebshr:
Glad to help — just started looking at this for an upcoming project and it looks really awesome!

PR #168, opened 2016-02-08
Fix crashing when reconnecting repeatedly

onevcat:
I am not quite sure why you are using `unowned` here. But this commit fixed some crash for me, when I try to reconnect to the web socket repeatedly while the network is bad.

I dealloc and reinit a web socket object every time the reconnecting button is clicked. And you may reproduce the crash easily if you turn off the network, and try to reconnect while you turn the network on.


daltoniam:
Ah ok. I just used unowned in those instances as I didn't think they would get called when an object was possibly deallocated or nil, but obviously that is not the case :smile: Thanks for the PR.

PR #174, opened 2016-02-24
Remove `var` parameters to conform to Swift 3

nicolasbigot:


daltoniam:
ah good point. Thanks!

PR #175, opened 2016-02-25
Set "Require Only App-Extension-Safe API" to YES

r-peck:
Without this setting, any application extensions will warn on linking with this framework and can result in app store rejection. Adding the setting only prevents use of APIs that are unavailable in the context of an extension (such as UIApplication.sharedApplication() on iOS).


daltoniam:
Thanks for the notes. I agreed it should be a problem since Starscream mainly uses foundation APIs. Thanks!

PR #176, opened 2016-02-27
Replace NSLock with pthread_mutex_t

noremac:
Speed!! :-)

### DIFF
--- a/Source/WebSocket.swift
+++ b/Source/WebSocket.swift
@@ -112,6 +112,7 @@ public class WebSocket : NSObject, NSStreamDelegate {
     public var security: SSLSecurity?
     public var enabledSSLCipherSuites: [SSLCipherSuite]?
     public var origin: String?
+    public var timeout = 5
     public var isConnected :Bool {
         return connected
     }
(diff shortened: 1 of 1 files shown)


nuclearace:
Using a spinlock is going to be pretty inefficient, unless the critical sections are really really fast. Also, I would be against adding a dependency that's only going to be available on Darwin platforms. Since it'd be another thing that would have to be changed for Linux support.

nuclearace:
Also e8dffea should be its own PR

acmacalister:
@noremac, Thanks for the PR. I appreciate you looking to contribute to the project. A few things to review:

1. As @nuclearace pointed out the spin lock can be inefficient depending on the context, so it would probably to be best to profile to see if the spinlock was an actual win.

2.  As far as Linux support goes, I would like to explore this more, but I agree @nuclearace that Foundation on Linux probably won't include an `OSSpinLock`. 

3. Upon further research an `OSSpinLock` might not be the best choice for our iOS platform and fanless MacBooks. See [here](https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2619) and [here](http://engineering.postmates.com/Spinlocks-Considered-Harmful-On-iOS/).

4. I also agree e8dffea should be in its PR. Let's discuss that one there.

5. I like 7e28ff5 to replace the force unwrapping. Let's do that in its own PR as well.

6. As far as swapping out `NSLock` was this because of a performance problem or just a potential optimization?

Any questions let me know. Thanks.

noremac:
I hadn't previously read that OSSpinLock was unsafe on iOS, that's good to know. OSSpinLock should be the fastest for cases where we aren't expecting to be waiting long (like the usage patterns in Starscream), but if it's not safe, it's not worth it :-) . Foundation on Linux is not something I'm in the habit of considering yet, and that's another good point.

For fun, I've benchmarked NSLock/OSSpinLock/pthread_mutex_t here: https://gist.github.com/noremac/cf57f662606a38cbce21

On my iPhone 6s, I get these results (nanoseconds)

nslock 54 nanoseconds
spinlock 30 nanoseconds
mutex 42 nanoseconds

If there's interest, I can go ahead and switch over to a mutex for further evaluation.

As for the other two commits, my mistake. I'll split them into separate pull requests.

noremac:
Alright, I've gone ahead and replaced NSLock with pthread_mutex_t.

PR #178, opened 2016-02-29
Adds an optional `writeCompletion` callback

noremac:
I often need to write files that I'm lazily reading out of an `NSInputStream`. The addition of a `writeCompletion` callback to the `write(data:)` method lets me smartly/lazily frame up the next chunk to send as soon as the socket is ready for more.

That's what I'd want it for anyway :-)

### DIFF
--- a/Source/WebSocket.swift
+++ b/Source/WebSocket.swift
@@ -180,23 +180,37 @@ public class WebSocket : NSObject, NSStreamDelegate {
         }
     }
     
-    ///write a string to the websocket. This sends it as a text frame.
-    public func writeString(str: String) {
+    /**
+     Write a string to the websocket. This sends it as a text frame.
+     
+     If you supply a non-nil completion block, I will perform it when the write completes.
+
+     - parameter str:        The string to write.
+     - parameter completion: The (optional) completion handler.
+     */
+    public func writeString(str: String, completion: (() -> ())? = nil) {
         guard isConnected else { return }
-        dequeueWrite(str.dataUsingEncoding(NSUTF8StringEncoding)!, code: .TextFrame)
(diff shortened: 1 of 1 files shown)


Source/WebSocket.swift
```diff
@@ -185,11 +185,18 @@ public class WebSocket : NSObject, NSStreamDelegate {
...
+
+    /**
+     Write binary data to the websocket. This sends it as a binary frame.
+     
+     If you supply a non-nil completion block, I will perform it when the write completes.
+
+     - parameter data:       The data to write.
+     - parameter completion: The (optional) completion handler.
+     */
+    public func writeData(data: NSData, completion: (() -> ())?) {
```
acmacalister: Can we set `completion` to default to nil? 

`writeCompletion: (() -> ())? = nil`

That way we can avoid can breaking changes for current versions that upgrade.

acmacalister:
Looks good! Other than that one comment, can we also add the completion closure to `writeString` and `writePing` to be consistent? Thanks!

noremac:
Oops, yes I had meant that to default to nil! I've updated writeString and writePing as well, good idea.

I did not update the comment for writePing as I'm not 100% sure what's going on there :-).

Thanks for the feedback.

acmacalister:
Awesome. I don't think `writePing` is all that commonly used and I would like clean up the rest of the docs anyway so I think we are good. Thanks again!

lhecker:
@acmacalister Do you think it's feasable to release a new version with this feature soon?

acmacalister:
@lhecker new release published (1.1.3) with this addition.

lhecker:
Thanks @acmacalister! This new feature is really useful to correctly handle backpressure. :blush:

PR #189, opened 2016-05-11
process buffered messages non-recursively

mayoff:
I had a Starscream crash because Starscream read a full buffer (BUFFER_MAX = 4096 bytes), and the buffer was full of 17 byte messages (meaning, 240 full messages plus most of the next message). The mutual recursion between `processRawMessage` and `processExtra` caused a stack overflow. (The default stack size for a background queue is 512 KB and, at least in a Debug build, the `processRawMessage` method uses a lot of stack space.)

This patch changes the buffer handling to be non-recursive.

The old `processRawMessage` method is now `processOneRawMessage`.  It returns the “extra” contents of the buffer instead of calling `processExtra` recursively.

The old `processExtra` method is now the `processRawMessagesInBuffer` method. It contains a loop that calls `processOneRawMessage` repeatedly until there's no more data in the buffer.

All calls to the old `processRawMessage` method (outside of `processRawMessagesInBuffer`) now call `processRawMessagesInBuffer`.

### DIFF
--- a/Source/WebSocket.swift
+++ b/Source/WebSocket.swift
@@ -408,7 +408,7 @@ public class WebSocket : NSObject, NSStreamDelegate {
         if !connected {
             processTCPHandshake(buffer, bufferLen: length)
         } else {
-            processRawMessage(buffer, bufferLen: length)
+            processRawMessagesInBuffer(buffer, bufferLen: length)
         }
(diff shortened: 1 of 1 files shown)


daltoniam:
As always, thank you for the excellent write up.

PR #202, opened 2016-06-04
Proxy connect

oldshuren:
This pull request added proxy support. The proxy info is from the system configuration. PAC is supported.

Proxy protocol supported include HTTPS (via CONNECT method) and SOCKS


nuclearace:
Style is inconsistent. Should probably be using type inference.

oldshuren:
Hi,

Could you be more specific?

Did you mean I should use delegate, not block? But the ProxyConnect is 
used only  once at the connection time, block is perfect for this situation.

Thanks,

Dong

Source/ProxyConnect.swift
```diff
@@ -0,0 +1,428 @@
...
+        }
+        if let proxySettings: NSDictionary = CFNetworkCopySystemProxySettings()?.takeRetainedValue() {
+            let proxies : NSArray = CFNetworkCopyProxiesForURL(hURL!, proxySettings).takeRetainedValue()
+            if proxies.count == 0 {
+                ProxyFastLog("configureProxy no proxy")
+                openConnection()
+                return
+            }
+            
+            let settings:NSDictionary = proxies[0] as! NSDictionary
```
nuclearace: Like here, you don't need to explicitly type since you're doing `as!`. You're also inconsistent with your `:`s. The style used should be `let x: Type`

oldshuren: Oh, I see.

I did some changes please take a look.

I just came from the objC side, still not very familiar of these finer 
points of Swift programming :(

Thanks,

Dong

daltoniam:
Thanks for this. There is quite a bit of code in here so it is going to take a bit to get through it all. At first glance there appears to be a tad bit of repeated code in the proxy class. Also I noticed you renamed a few of the method names in the main WebSocket class. I will hopefully give some more detail feedback in the coming days. Thanks again!

daltoniam:
Thanks for the explanation. I appreciate knowing the history and can see the port of the SocketRocket implementation. A few high level things I would like to see:

1. Design to be more inline with Starscream current architecture. I understand the use case of the separate class for proxy connections, but this feels a lot like it is a port of the SocketRocket code vs being designed with Starscream's architecture. SocketRocket and Starscream are designed and function differently and I would like to keep the current architecture intact. I understand your point with needing to open a connection to the proxy server then to the WebSocket server, but I think it would be a relatively minor change to add the proxy support into the current handshake process instead of having to move half of the logic into the proxy class. It seems strange that the proxy class is partially responsible for checking a proxy request regardless if a proxy is being used or not.

2. Idiomatic Swift. Not to be overly stylistic or harp on the objective-c port, but this has a couple of style things I would like to see changes, such as a few unneeded semi colons, remove the logging, a few properties that might be able to use a `guard` or unwrapped optionals instead of the explicit unwrapping.

3. This doesn't have anything to do with the code, but we will need to test that it still passes the autobahn tests with both a proxy and not a proxy. If you have any recommendations on a good test proxy, please share.

I want to stress again I appreciate the Pull Request and for you taking the time to put this together. Any questions please let me know.

oldshuren:
Hi,

I submitted pull request to the socket.io-client-swift and they tell me 
to submit pull request to you.

In that pull request, the proxy handling is actually done in the 
WebSocket.swift class. Here is the link,

https://github.com/oldshuren/socket.io-client-swift/blob/proxy-connect/Source/WebSocket.swift

Do you think it is better this way? I can clean it up on the style part.

Thanks,

Dong

daltoniam:
Thanks for the reference. Yes, this appears to be more inline with the design structure.

oldshuren:
I merged the proxy code back to WebSocket.swift now.

Dong

daltoniam:
Did a quick glance I like this structure much better. I will try and give a more feedback soon. Thanks again!

hifall:
Is there any further progress on this matter? It'd be fantastic for proxy support to be added.

jariz:
I'm also interested in an update from @daltoniam.

daltoniam:
This one has taken me a while due to personal time restraints and the size of changes. It also requires testing with and without a proxy server behind an autobahn server which is a fair amount of setup for a test. I am hoping to look at this weekend (along with a proper Swift 3 and Swift 2.3 tagged releases).

vanvalkenburg:
Is this going any further? It would be great to get proxy support added. I have a requirement to support proxy settings, and I'd assume this is fairly common in enterprise environments.

theinfamousmielie:
@daltoniam is anything happening with this? I'm really looking forward to this change...

JoeSzymanski:
Is there any possibility of getting this functional and merged? Testing our product would be significantly easier if we could proxy the connections through Charles and see the content.

daltoniam:
Yes, I still would like to see proxy support add to Starscream, but I still just haven't had the time or willpower to do this. If anyone else wants to take what is here and address said comments in a PR it would amazing. You would also have my undying appreciation 😄.

Shadester:
What is the said comments that needs to be addressed? 1 seems to be addressed? And maybe 2, too?

ergunkocak:
Hello,

First of all tnx a lot for such library. 
Proxy support is kind a very important to inspect the traffic. And in my current project not having proxy support is really crippling me. If there is anything i can do please explain more in detail so i can put my best to help.

Best regards

OneSman7:
I am also waiting for this to be merged. Will be glad to help if you say what do you need.

mbesnili:
This is a key feature, anyone to merge this ?

daltoniam:
I would like to see this get merged, but all 3 of the initial points still stand.
 
1. This needs to be update for the 3.0.0 refactor. It might be easier to have a less intrusive proxy design into the WebSocket handling then previously.
2. the style needs to be stylish swift (bunch of unnecessary semi colons, etc).
3. This is just a testing angle, but an important one since this change will require a way to have a proxy server and an autobahn server. The Starscream autobahn test project will have to be updated to support testing this new case.

gjsbe:
Hi,
on ios platform we need https to call rest webservices, and websockets are served via wss.
When there's a proxy server in front of an internet connection, often is an http/https proxy and not a socks proxy. Squid for example is an http/https proxy but doesn't support socks.
wss is an upgrade of https connections and usually comes with https protocol.
Starscream supports socks but not https tunneled proxy connection, I need it for a project.
I see these feature, what do you need to implement it? Can I use it in my code anyway?
Thank you very much
Regards

francksyren:
Awesome library, please merge this pull request

oinsider:
In my case, if I run the following code (CFNetworkCopyProxiesForURL) taken from the @oldshuren pull request I don't see any proxies. 
In the list there is only one element [kCFProxyTypeKey: kCFProxyTypeNone]

```
        if let proxySettings: NSDictionary = CFNetworkCopySystemProxySettings()?.takeRetainedValue() {
            let url = "wss://endpoint.socket.io"
            let nsURL = NSURL(string: url)!
            let proxies: NSArray = CFNetworkCopyProxiesForURL(nsURL as CFURL, proxySettings).takeRetainedValue()
            print(proxies)
        }
```

I did notice that going directly to the machine without proxies I get the response:
HTTP/1.1 101 Switching Protocols
Upgrade: websocket
Connection: Upgrade

Otherwise if I go through the ELB and HAProxy I get:
HTTP/1.0 404 Not Found
Cache-Control: no-cache
Connection: close
Content-Type: text/html
 
<html><body><h1>404 Not Found</h1>
</body></html>

It seems that the websocket are connecting, but for some reason, the response to the Get request that is sent is resolved or not.

gjsbe:
Hi,
this is the problem: if you ask any os (android or ios) how to reach wss://something it tells you DIRECT.
I think because wss protocol isn't recognized by os api.
You have to ask for https://something (for wss), then use an https proxy with connect, but with this version of starscream won't work via https proxy.

oinsider:
I am trying to get the proxy list with https in the urls and even with that CFNetworkCopyProxiesForURL doesn't notice any proxy. If this function doesn't get any proxy no configuration is going to be applied to the socket connections.

gjsbe:
But did you configure your wifi connection to use proxy?
Because it works for me...

oinsider:
Yes, if I route the traffic to the charles proxy I can see the proxy in the list. But in my case the proxy is behind the domain.

gjsbe:
mmm... I don't know, I think you have a configuration problem, because CFNetworkCopyProxiesForURL  is not starscream related, is Core Foundation function.

oinsider:
CFNetworkCopyProxiesForURL is used by the proxy-connect pull request in order to get the possible proxy types. These can be kCFProxyTypeAutoConfigurationURL or kCFProxyTypeAutoConfigurationJavaScript. 

In the first case is going to get the PAC proxy configuration file and read the proxies of the PAC file. In the second case, it just reads the proxies of the PAC file.

Then when reading the proxy settings it gets the httpProxyHost and httpProxyPort if the proxy type is HTTP/HTTPS or the socksProxyHost, socksProxyPort, socksProxyUsername and the socksProxyPassword if the proxy type is SOCKS. 
Then inits the read and write streams with the configurations of the proxies. 

Maybe I should try to configure the PAC file in the proxy and then check if the CFNetworkCopyProxiesForURL notices it.

oinsider:
Issue fixed for me, there was a problem in the haproxy config.

acmacalister:
The codebase has actually under gone a major refactor which has `Network.framework` support. Since that is able to handle proxy connections transparently, I'm going to close this for now. Feel free to check it out here #653.

PR #204, opened 2016-06-14
Let SSLCert and SSLSecurity inherit from NSObject

nuclearace:
This let's them be accessed from Objective-C.


acmacalister:
hey @nuclearace. What is the use case for needing these objects to be accessed from Objective-C? Historically, we have all the Objective-C compliant stuff be done in the separate JetFire project.

nuclearace:
https://github.com/socketio/socket.io-client-swift/issues/395 is my need. It's not a huge change, I just thought someone else might run into a similar case.

PR #205, opened 2016-06-14
Xcode 8 support

soffes:
Guessing you don't want to merge this yet, but thought I'd PR anyway so you have it if you want it down the road. Thanks for all of your work on this!

### DIFF
--- a/Source/SSLSecurity.swift
+++ b/Source/SSLSecurity.swift
@@ -139,9 +139,9 @@ public class SSLSecurity {
         }
         var policy: SecPolicyRef
         if self.validatedDN {
-            policy = SecPolicyCreateSSL(true, domain)
+            policy = SecPolicyCreateSSL(true, domain)!
         } else {
-            policy = SecPolicyCreateBasicX509()
+            policy = SecPolicyCreateBasicX509()!
         }
         SecTrustSetPolicies(trust,policy)
         if self.usePublicKeys {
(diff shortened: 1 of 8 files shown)


Source/SSLSecurity.swift
```diff
@@ -139,9 +139,9 @@ public class SSLSecurity {
         }
         var policy: SecPolicyRef
         if self.validatedDN {
-            policy = SecPolicyCreateSSL(true, domain)
+            policy = SecPolicyCreateSSL(true, domain)!
```
soffes: I feel bad about this, but wanted to get it going quickly.

daltoniam:
Thanks! Agreed, it should probably be in its own branch for now. I started working on a Swift 3 version here: 
https://github.com/daltoniam/Starscream/tree/swift3
That was before xCode 8 was out, just a nightly release build. I think there will need to be a few breaking changes to be inline with the Swift 3 changes (such as `writeString` and `writeData` just being `write` because of the first parameter now being include.) I'll make sure to cross reference your changes to see if I missed anything with the official swift 3/xCode 8. Thanks again.

soffes:
@daltoniam no worries!

I made good progress here the other day: https://github.com/usecanvas/Starscream/tree/swift3

daltoniam:
Ah very nice! I like those changes, would like to pull them when they are ready. We are getting much closer to a GM release so I am hoping to take some time soon to solidify everything for the Swift 3 release.

soffes:
There's only 2 issues left. Both are related to pointer math. All of that is a little scary to me, so I'll let you take it from here if you want.

daltoniam:
Fair enough. Thanks!

PR #207, opened 2016-06-18
Support Swift 3.0 and Xcode 8.0 (8S128d)

ShinCurry:
I only converted syntax to Swift 3.0, uncertain whether there will be a new bug,

but it seems work well on my own project.


daltoniam:
Thank you for the PR. I had a swift 3 branch here:

https://github.com/daltoniam/Starscream/tree/swift3

I merged those changes with the xCode 8 converter changes as well (will bring some breaking changes). Thanks again!

PR #209, opened 2016-06-24
Returned custom close reason from server according to RFC 6455

Abeansits:
It appears to me as if this feature has been planned but not fully implemented. 
By actually using the parsed string we support close reason mentioned here: https://tools.ietf.org/html/rfc6455#section-7.1.6


daltoniam:
Ah nice catch. Thank you!

PR #214, opened 2016-07-06
Xcode 8 beta 2 updates

noremac:
Xcode 8 beta 2 has introduced a few new issues that needed to be addressed.

As the current swift 3 branch was a bit behind master, I've started fresh from the latest master and converted everything over again. Where Xcode's conversion failed, I've copied the existing fixes that were on the swift3 branch.

I'm not 100% sure if any other non-swift-3 changes have gone on here, so this may have not been the best approach. Let me know if you'd like an updated PR!


daltoniam:
Thanks for the PR. Can we point it at the Swift 3 branch? I know it is a tad behind, but shouldn't be anything major and couldn't easily be pulled in. Also the current Swift 3 branch brought in a couple of idiomatic changes for Swift 3 that I would like to keep (such as changing `writeString` and `writeData` to just `write`, with a named first parameter as required by Swift now). If we could include those changes and pointed to be merged into the Swift 3 branch I would appreciate it. Thanks!

noremac:
I could not figure out how to edit this PR appropriately, so I've made a new one with the correct destination branch here: https://github.com/daltoniam/Starscream/pull/218. The `write` methods are re-named as well.

PR #219, opened 2016-07-12
Fixing 'queue' naming error

nathanborror:


daltoniam:
Ah thank you much. I totally overlooked that.

daltoniam:
I'll get a CI setup at some point 😄

PR #222, opened 2016-07-21
Fix typo in `dequeueInput`

Danappelxx:
This was breaking tcp messages that were sent in chunks, since only the first chunk would be processed.

As you can imagine, this one took quite a while to figure out... (in my case, I could only reproduce the error on a local network and not over a tcp tunnel, so I totally misdiagnosed it at first).


daltoniam:
ah, apologize on the time it took to track this down, can't believe I did that. Thank you much!

Danappelxx:
No problem, got to explore the codebase 😃

PR #225, opened 2016-07-31
Fix typo that prevents SSL detection

ThinkChaos:


daltoniam:
Ah totally missed that. Thank you!

PR #227, opened 2016-08-02
Updated for Xcode 8 Beta 4

treyrich:


Source/WebSocket.swift
```diff
@@ -129,15 +129,15 @@ public class WebSocket : NSObject, StreamDelegate {
...
+    private let mutex = NSLock()
     private var canDispatch: Bool {
         mutex.lock()
         let canWork = readyToWrite
         mutex.unlock()
         return canWork
     }
     //the shared processing queue used for all websocket
-    private static let sharedWorkQueue = DispatchQueue(label: "com.vluxe.starscream.websocket", attributes: .serial)
+    private static let sharedWorkQueue = DispatchQueue(label: "com.vluxe.starscream.websocket", attributes: [])
```
nuclearace: Is this a bug they introduced? Why would they remove the .serial option?

treyrich: From what I understand Serial is default, so there's no need to specify.

daltoniam:
Appreciate the help, thanks!

Issue #229, opened 2016-08-09
Support of custom close codes and access to close message payload


serieuxchat:
I was wondering whether it would be possible to support the following:
- Ability to specify a custom code when closing the web socket (currently the disconnect method always sends CloseCode.Normal.rawValue)
- Ability to get the payload of the close message that is sent by the remote end.
  (Currently the private method processTCPHandshake only passes the received HTTP code to doDisconnect and the eventual callback, but there is no way to get the payload that came with the close message).

(The RFC refers to the possible custom body of the Close message in https://tools.ietf.org/html/rfc6455#section-5.5.1)

daltoniam:
- I figure this could be added, would just have to a small change to the disconnect method to support different codes (maybe do a default parameter or something).
- This is supported right now here: https://github.com/daltoniam/Starscream/blob/master/Source/WebSocket.swift#L619. Are you wanting to get the HTTP payload back on a failed upgrade request from `processTCPHandshake`?

serieuxchat:
(1) Would be really great if it's supported.
(2) Wow, that awesome! I guess, the reason string (UTF8 encoded data from section 5.5.1) would be enough for our purposes. I will give it a try asap.

daltoniam:
This was added in the 2.0.0 release (Swift 3 support)!

serieuxchat:
I was wondering whether it would be possible to have those in Swift 2.3 code as well?

commjoen:
+1 for the swift 2.3 code as we are using this in our project :).

PR #237, opened 2016-08-22
Support a data that may be passed in PONG from the server

nonameplum:
According to the https://github.com/daltoniam/Starscream/issues/229 added optional `data` parameter that may be filled by a server with `PONG`.


daltoniam:
I'm fine with this change, but could I request it be pointed at the swift 3 branch? It is a breaking change, so better to be included in that release. Also is there a use case you can share for this? I haven't see a setup that requires the PONG data on the client side (not saying there isn't, just have personally seen it before).

nonameplum:
Ok, I will do this. I created app where server only kept web socket connection when client was sending ping repetitively. Pong message was used to store connection latency (pong data has timestamp when ping was requested and id).

daltoniam:
I ended up just manually putting these changes into the Swift 3 branch. Thanks!

PR #238, opened 2016-08-22
Added optional convenience callback closure parameter to the connect …

nonameplum:
…method that will be called when the client connects to the server.


nuclearace:
What's wrong with the onConnect handler?

nonameplum:
Nothing. On Connect is in different place and you have deal with that. Here at the time of invoking connection you can pass a closure.

daltoniam:
I think I agree with @nuclearace on this one. The `onConnect` handles this case. I understand the difference with this only firing once, but that could easily be accomplish with setting the onConnect closure to nil after the first connect to. Thank you for the PR though and taking the time to send it in.

PR #242, opened 2016-08-28
Xcode8b6

pvzig:
Updates to allow the project to be built with Carthage with the Xcode 8 beta 6 tooling:
 
- Set `SWIFT_VERSION` to `3.0` for all targets
- Update sample test syntax so it compiles


acmacalister:
Thanks!

PR #243, opened 2016-08-29
Added initializer to set the QOS of the write queue

cpinski:
Changes:
* Added an initializer to set the QOS of the write queue. We noticed in our project that the write queue was getting too backed up which would cause the app to not respond to the hardware quickly enough. This will allow the app side to decide whether it wants to override the default in case a higher or lower QOS is needed.


daltoniam:
I like it. Thanks!

PR #253, opened 2016-09-25
Only disable SSL cert validation if we are using a secure scheme.

noremac:
Enabling `disableSSLCertValidation` against a non-secure scheme (`"ws"`) can cause some slightly hard to track down connection failures. This PR fixes that issue by only checking and handling `disableSSLCertValidation` in an area where we know we are using a secured scheme (`"wss"` or `"https"`).

Additionally, we may want to log this mismatch with a simple print statement as well?

I could see this as an area where no user-error is tolerated (figure out if you are secure or not _darn it!_ 😄) so, some might see this PR as slightly dangerous. I think it's useful though.

(We may want to add the `if let cipherSuites...` line up into this same area as well?

Let me know what you think!

### DIFF
--- a/Source/WebSocket.swift
+++ b/Source/WebSocket.swift
@@ -325,34 +325,34 @@ public class WebSocket : NSObject, StreamDelegate {
         if supportedSSLSchemes.contains(url.scheme!) {
             inStream.setProperty(StreamSocketSecurityLevel.negotiatedSSL as AnyObject, forKey: Stream.PropertyKey.socketSecurityLevelKey)
             outStream.setProperty(StreamSocketSecurityLevel.negotiatedSSL as AnyObject, forKey: Stream.PropertyKey.socketSecurityLevelKey)
+            if disableSSLCertValidation {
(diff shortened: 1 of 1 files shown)


daltoniam:
Thanks for the PR and write up. Just to make sure I understand this correctly, if you use `disableSSLCertValidation` on a non secure connection, it will cause the connection to fail with an arbitrary error?

noremac:
Correct! It looks like the underlying streams get confused when kCFStreamPropertySSLSettings are set, even if you aren't trying to negotiate an SSL connection.

daltoniam:
Ah good deal. Yeah that seems to make sense to do it for the cipher suites as well. Go for it and thanks!

noremac:
All set! Let me know if there's anything else.

daltoniam:
Sweet! One question though, wouldn't we want the `cipherSuites` to be outside of the `disableSSLCertValidation`? That would only be usable if you turned off SSL verification?

noremac:
If I'm understanding this correctly, it seems like that's what I've done. The `enabledSSLCipherSuites` check is not nested within the `disableSSLCertValidation` check. Is that what you meant?

daltoniam:
Ah my bad, I misread it. Looks good, thanks!

PR #255, opened 2016-09-25
Update OS X deployment target

pvzig:
`NSQualityOfService` requires 10.10 or later.


daltoniam:
Ah forgot to do it for the Swift 2.3 branch. Thanks!

PR #258, opened 2016-09-27
Make it work on Xcode 7.3.1

jcesarmobile:
added some `#if swift(>=2.3)` to make it work on Xcode 7.3.1

### DIFF
--- a/Source/Info-tvOS.plist
+++ b/Source/Info-tvOS.plist
@@ -15,7 +15,7 @@
 	<key>CFBundlePackageType</key>
 	<string>FMWK</string>
 	<key>CFBundleShortVersionString</key>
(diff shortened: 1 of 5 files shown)


nuclearace:
I don't think this is appropriate, if you're using the swift2.3 branch, I think you should be on Xcode 8.

corinnekrych:
Swift 2.3 is also compatible in Xcode 7.3.1
I found it nice to be able to build it with both, but since Xcode8 is now released it might be less important.

jcesarmobile:
@nuclearace I thought Xcode 7.3.1 was compatible with Swift 2.3, but you are right, it isn't.

Closing the PR

PR #259, opened 2016-09-28
Improve clarity of asyncAfter code and other timeout

robinkunde:
Hi, I'd like to improve the clarity of the deadline calculation for Dispatch.asyncAfter in WebSocket.disconnect(). I think the existing code is a result of the Swift3 code updater and could use some syntactic sugar.

I also added underscores to time-related constants to improve legibility.

### DIFF
--- a/Source/WebSocket.swift
+++ b/Source/WebSocket.swift
@@ -201,7 +201,8 @@ public class WebSocket : NSObject, StreamDelegate {
     public func disconnect(forceTimeout: TimeInterval? = nil, closeCode: UInt16 = CloseCode.normal.rawValue) {
         switch forceTimeout {
         case .some(let seconds) where seconds > 0:
-            callbackQueue.asyncAfter(deadline: DispatchTime.now() + Double(Int64(seconds * Double(NSEC_PER_SEC))) / Double(NSEC_PER_SEC)) { [weak self] in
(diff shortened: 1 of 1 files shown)


Source/WebSocket.swift
```diff
@@ -201,7 +201,8 @@ public class WebSocket : NSObject, StreamDelegate {
     public func disconnect(forceTimeout: TimeInterval? = nil, closeCode: UInt16 = CloseCode.normal.rawValue) {
         switch forceTimeout {
         case .some(let seconds) where seconds > 0:
-            callbackQueue.asyncAfter(deadline: DispatchTime.now() + Double(Int64(seconds * Double(NSEC_PER_SEC))) / Double(NSEC_PER_SEC)) { [weak self] in
+            let milliseconds = Int(seconds * 1_000)
+            callbackQueue.asyncAfter(deadline: .now() + .milliseconds(milliseconds)) { [weak self] in
```
max-potapov: You already have seconds to use `now() + .seconds(Int(x))` function.

robinkunde: That would remove the ability to specify fractional timeouts like 2.5 seconds. That's why I'm doing the conversion to milliseconds first.

daltoniam:
It was and I like the clarity changes. Thanks!

PR #266, opened 2016-10-10
'Open'ing classes

### LINKED ISSUE #265: WebSocket class should be 'open' not 'public'
Defining classes with `public` access control keeps them from being "subclassable" outside its defining module. The behaviour is similar to declaring a class as `final`.

Therefore, WebSocket should be `open` which means it is public and inheritable.

fjcaetano:
fix #265

### DIFF
--- a/Source/SSLSecurity.swift
+++ b/Source/SSLSecurity.swift
@@ -23,7 +23,7 @@
 import Foundation
 import Security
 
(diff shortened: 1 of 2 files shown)


daltoniam:
Thanks! Not sure how I overlooked that (even with the migrator).

Issue #257, opened 2016-09-27
Strange crash


brightsider:
https://github.com/socketio/socket.io-client-swift/pull/515#issuecomment-249829170
Can you make update for this issue?

daltoniam:
Can you provide anymore information than just a screenshot? I'm not sure how to reproduce this behavior without some more info. Thanks!

Also to have it for reference.

![](https://cloud.githubusercontent.com/assets/4996231/18663280/810068ba-7f26-11e6-8b92-5a27f06deaaf.png)

brightsider:
I don't understand why it's crashed, but pull helps me.
Why make as to AnyObject, then to SecTrust, if can only as SecTrust? This helps me :)

nuclearace:
@daltoniam His fix for this was https://github.com/socketio/socket.io-client-swift/pull/515/files, I said he should reopen the PR here.

daltoniam:
2.0.1 with this fixed released!

PR #267, opened 2016-10-10
[Fixup] crash for SSL Pinning Strange crash

Dipak99041012:
There is strange crash when it does SSL Pinning 
issue - 257
https://github.com/daltoniam/Starscream/issues/257


daltoniam:
Thanks!

PR #268, opened 2016-10-14
Add callbacks for convenience.

SkylerLutz:
## Description

In the interest of testability, I found that hanging hooks off `WebSocket` was an effective means of asserting the appropriate ping-pong behavior. @daltoniam @chitowncrispy please review


acmacalister:
@SkylerLutz sorry for taking so long to get back to this. We aren't adding anything to Swift 2.3 anymore as Apple is swift 2.3 deprecating very shortly and #298 is the same thing for Swift 3.

http://ericasadun.com/2016/11/10/xcode-8-2-deprecates-swift-2-3/

Thanks for the PR though!

PR #276, opened 2016-10-24
Enable custom trust validation

profer:
In my project I need to perform special trust validation that is not covered by the existing SSLSecurity class. I saw that pull Request #266 opened all classes for subclassing but does not allow overriding of any of the methods.
So I change `WebSocket` to use a new protocol - `SSLTrustValidator` - for trust validation instead of `SSLSecurity` directly. This allows users of Starscream to supply their own validation logic if needed.

### DIFF
--- a/Source/SSLSecurity.swift
+++ b/Source/SSLSecurity.swift
@@ -23,6 +23,10 @@
 import Foundation
 import Security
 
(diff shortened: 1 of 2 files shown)


daltoniam:
Make sense, I like it. Thanks!

Issue #264, opened 2016-10-10
Errors trying to download using Swift Package Manager


MarcusSmith:
When trying to do a swift build with Starscream in my Package dependencies I'm seeing this error:

```
error: the package has an unsupported layout, unexpected source file(s) found: .../Packages/Starscream-2.0.0/Tests/StarscreamTests.swift
fix: move the file(s) inside a module
```

daltoniam:
Hrm, not sure. I'll have to checkout the package manager again, haven't touched it since Swift 2.X. 

@nuclearace maybe I am misremember, but aren't you a Mac/Swift Package Manager user? I'll look into it to, just didn't know if you had any ideas off hand.

nuclearace:
Yeah I saw this when working on my new Discord library. Seemed to happen on the latest preview release of swift and the pm. Didn't make much sense to me deleting the test folder and rebuilding is a workaround

nuclearace:
I think you need to move the file in Tests into its own folder. Like `Tests/StarscreamTests/*.swift`

daltoniam:
#277 should have fixed this (in master now).

abrain:
Thanks for resolving this. It seems Swift Package Manager still picks tag 2.0.1, so the error continues to exist until a new release.

daltoniam:
`2.0.2` released.

PR #282, opened 2016-11-12
Fix tests

hug33k:
Fix to avoid this error :
```
error: the package has an unsupported layout, unexpected source file(s) found: $PROJECT/Packages/Starscream-2.0.1/Tests/StarscreamTests.swift
fix: move the file(s) inside a module
```


nuclearace:
Should already be fixed in master but a new version needs to be released

daltoniam:
@nuclearace is correct, but thank you for the PR. The `2.0.2` release resolved this. Also see #264 for more info.

Issue #287, opened 2016-11-19
Origin should not include path information


dancasimiro:
From the Mozilla documents:

The origin is a URI indicating the server from which the request initiated.  It does not include any path information, but only the server name.

I think that sending the absolute URL is violating the spec. Some servers reject an origin with path information as malformed.

daltoniam:
#288 is merged. Thanks again!

PR #288, opened 2016-11-21
Remove the path components from the URL sent as origin

dancasimiro:
This fixes issue #287. The origin is a URI indicating the server from which the request initiated. It does not include any path information, but only the server name.


daltoniam:
Thanks! Sorry about the delay. Holidays make it tough to get to this stuff 😄.

PR #292, opened 2016-12-19
Change public to open

euskadi31:


daltoniam:
I'm fine with this change, but out of curiosity, is there a case where you need to subclass the WebSocket object?

euskadi31:
For unit test (mock)

Envoyé de mon iPhone

PR #294, opened 2016-12-27
Prevent reconnection crash

### LINKED ISSUE #272: Establishing new connections connections causes app to crash over time
We're having an issue which causes the application to crash after we've connected several times (say 50+) which causes the application to crash.

We are wondering if it's something to do with using WSS due to seeing this commit:
https://github.com/daltoniam/Starscream/commit/20285cce972762929a497abf616b1a460b499f32

Here is a link to the crash:
https://gist.github.com/darrenhaken/7a9c2adc772de12e9cd50a1ba725fb92

koke:
There are a couple cases where a socket would create a new connection without closing an existing one.

1. Between the time when the socket opens the streams until the TCP handshake ends, it would be both `!connected` and `!isConnecting`, so a second `connect` would go through creating a new pair of streams without disconnecting the first pair.
2. Since `disconnect` without a timeout only sends a close message to the server, if we tried to `connect` the socket again before the connection was closed it would not disconnect the previous streams.

In any of those cases we end up with a `NSInputStream` that still has the `WebSocket` as a delegate, but it's no longer referenced in `inputStream`. If the `WebSocket` is deallocated, it wouldn't clean up the delegate on that stream, and if it received any data, it would try to forward it to a deallocated object, which would crash.

This should fix #272

### DIFF
--- a/Source/WebSocket.swift
+++ b/Source/WebSocket.swift
@@ -190,7 +190,6 @@ open class WebSocket : NSObject, StreamDelegate {
         didDisconnect = false
         isConnecting = true
         createHTTPRequest()
-        isConnecting = false
     }
 
     /**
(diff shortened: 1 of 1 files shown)


daltoniam:
Ah very good. Thank you for the great write up and PR!

PR #296, opened 2016-12-31
Fix #295

### LINKED ISSUE #295: Creating a new client blocks disconnects from server.
It looks like 14313df360c12bfa0d760c0bffb9ccbab16f6c43 is causing an issue where the client goes through the motions of shutting itself down in [`doDisconnect`](https://github.com/daltoniam/Starscream/blob/master/Source/WebSocket.swift#L911), but stops short of notifying that it disconnected.

This causes an issue because subsequent calls to `doDisconnect` won't notify that it's actually disconnected because there's a guard against trying to call this twice.

nuclearace:
@daltoniam Please review.

### DIFF
--- a/Source/WebSocket.swift
+++ b/Source/WebSocket.swift
@@ -500,6 +500,7 @@ open class WebSocket : NSObject, StreamDelegate {
         case 0:
             isConnecting = false
             connected = true
+            didDisconnect = false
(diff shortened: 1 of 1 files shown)


daltoniam:
Ah thanks! I'm going to merge this and run through some testing. Thanks again.

Issue #297, opened 2017-01-03
How do you run the tests?


Elethier:
My team and I have been working on a project using Starscream, and we've had to make some minor modifications for our use case, as well as some bug fixes. We were hoping to submit some of those changes for a pull request. I wanted to run the tests before submitting, but I'm having trouble getting the tests to pass, even before applying the changes we've made. Can you explain how to run the test suite for this project?

I noticed the project is using the Autobahn test suite, so I set that up and ran the AutobahnTest project. However, I'm noticing some failures on test cases, so I'm not sure what I'm doing wrong. There was no information on what mode to run the `wstest` command provided by the Autobahn test suite, so I tried `wstest -m fuzzingserver` and it seemed to work: some, but not all, of the test cases pass. Am I doing something wrong? Are there some outstanding issues with getting the library to pass all of the Autobahn test cases? 

Thanks!

acmacalister:
What test are failing? Running the fuzzing server is correct. Every once and while, some of the test will fail at random. Maybe like one or two, but it is never consistent which ones will fail, if any. Have you looked at `AutobahnTest` in the examples folder?

Elethier:
@acmacalister I ran the fuzzing server and AutobahnTest twice. The first run, I saw 249 failures out of 519 cases. The second time, I saw 250. They're not the same tests. Is this consistent with what you're seeing on your end? (I'm guessing it's not...)

daltoniam:
Ah, this is our fault with the logging in the Autobahn tester app. The results we check are under client reports at `http://localhost:8080`. The logging on that test app wasn't finished to handle all the cases so it will print failures for tests that aren't implemented (the compression ones of 12 & 13).

TLDR; 
check http://localhost:8080 for results instead of this print in the autobahn tester app 😄.

PR #298, opened 2017-01-03
Added callbacks for pings and pongs

Elethier:
I took the changes from #268 and reimplemented them in Swift 3. These changes add callbacks and a delegate for apps to listen for when a ping has been sent, a ping has been received, and a pong has been sent.

@daltoniam Please review.


acmacalister:
@Elethier more of an exploration question. I understand this is used for testing, but why do you need to ping/pong the server in the first place? Most WebSocket libraries (this one included) handle the ping/pong behavior for you...

Elethier:
@acmacalister For my project, we're using these callbacks for purely diagnostic purposes. The code has been invaluable in diagnosing our connection issues with the hardware we're trying to connect to. I was able to check what time we were sending pongs, the exact time a ping was sent down the OutputStream, etc.

We are _not_ using the callbacks to implement ponging behavior, since that's already implemented within the library. I can see how this could cause confusion, though. Maybe I should add some documentation. I'm open to suggestions.

cpinski:
@Elethier can we add some documentation as to how these callback would be used? I think that will help other developers understand the meaning for these callbacks and correctly use them.

Elethier:
@cpinski I added some documentation. Although some of it looks a little redundant...please comment if you think there's some tags that could be removed.

Elethier:
@daltoniam @acmacalister What do you two think? Does this PR seem suitable to merge in?

daltoniam:
I see the value in debugging with this, but I fear it would cause confusion from a implementation stand point and its main value is in debugging. It doesn't inherently add much to a normal user outside of being able to debug which is hopefully not that often 😁.  Also the `onPong` was change to a suitable name, but that will cause some minor issues with backwards compatibility (not a big deal, just as a note). I guess my concern is this doesn't add much value to the typical user of the library and is only really handy for those of us that work on the code base directly to debug with. Nothing wrong with the code, just more a question if it adds enough value and if there is a way we might be able to make it less intrusive. I'm certainly open to suggestions, but I haven't thought of a way yet to accomplish this yet.

Elethier:
@daltoniam Perhaps I could add something like a `WebSocketDebugDelegate` on top of the existing functionality? I'm worried that if someone else needs to do the debugging I did, they'll have to fork the library and reimplement these changes. Then again, we'd be forcing people to include debug code into production builds...unless I did an `#if DEBUG` inside of the library.

Elethier:
@daltoniam I wanted to bring this up again, since this pull request has been up for a while. Do you think it would be valuable to merge this in? Do you think a `WebSocketDebugDelegate` would be a better approach? Or should I just close this pull request? 

Since implementing these changes in our project, we haven't had any major connection issues, so I wouldn't mind removing the related code in our project and closing this.

jakeholland:
@daltoniam I will be taking over this PR for @Elethier from this point forward. Just curious if you had any open questions?

hishnash:
This would be nice, would work well along side the features in https://github.com/daltoniam/Starscream/pull/337

acmacalister:
@jakeholland @Elethier apologies on the massive delay here. I like @hishnash idea of adding this to the `AdvancedWebSocketDelegate`. I'm going to merge the other PR tomorrow, would you mind updating this PR to use that instead? Again apologies on the delay.

jakeholland:
I should be able to get to this tomorrow I think.

daltoniam:
@jakeholland again sorry about the delay, but finally merged 😄. One of these days I will be fast at merging PRs. Thanks again.

jakeholland:
@daltoniam - I  finally got around to making the changes to move the ping/pong callbacks into the new AdvancedWebSocketDelegate 😃

daltoniam:
@jakeholland Thanks! No worries, I'm not exactly the fastest on open source anymore either 😬. I'm going to create a Swift 4 branch and merge these changes into that since it does have some minor backwards compatibility breaking changes with the pong stuff.

zayats77:
Hi
Please merge this pull request.

PR #300, opened 2017-01-04
Added guard against multiple disconnects

Elethier:
I was noticing issues where my app would send multiple disconnect frames. It seemed appropriate for the library to guard against this. (Although there's an argument to be made that avoiding unnecessary calls to disconnect is the responsibility of whomever is using the library...)

@daltoniam Please review.


daltoniam:
Thanks! Seems fair as all the other `write` methods protect against it.

PR #302, opened 2017-01-04
Fixed issues with not cancelling block operations properly

Elethier:
I was noticing some issues where the library sometimes spawns a lot of extra threads (around 100 threads). I tracked down the issue to the way the library uses BlockOperation. While the library calls code calls `cancelAllOperations`, the code within the block operations doesn't check the isCancelled flag during execution. This is incorrect, based on Apple's documentation on [OperationQueue](https://developer.apple.com/reference/foundation/operationqueue): 

> Canceling an operation object leaves the object in the queue but notifies the object that it should abort its task as quickly as possible. For currently executing operations, this means that the operation object’s work code must check the cancellation state, stop what it is doing, and mark itself as finished. 

As a result, the block operations weren't cancelling properly. To fix the issue, I revised the use of BlockOperation so that the code checks the isCancelled flag at appropriate times, such as in `while` loops. 

@daltoniam Please review.

### DIFF
--- a/Source/WebSocket.swift
+++ b/Source/WebSocket.swift
@@ -372,9 +372,12 @@ open class WebSocket : NSObject, StreamDelegate {
         
         let bytes = UnsafeRawPointer((data as NSData).bytes).assumingMemoryBound(to: UInt8.self)
         var out = timeout * 1_000_000 // wait 5 seconds before giving up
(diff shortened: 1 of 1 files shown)


daltoniam:
Nice write up! Thanks for this. I ran it through the tests as well and all appears well. Thanks again.

PR #311, opened 2017-02-20
Added WebSocketClient protocol for mocking

euskadi31:


euskadi31:
up

daltoniam:
Sorry about the delay in responding. I have been going back and forth on this. I get the usefulness for mocking in creating the `WebSocketClient` protocol, but it does add a little bit of complexity just for the sake of mocking. I'm kinda of indecisive on if that is a trade off worth making. Not to drag this out any longer, but let me get some feedback from some folk. Thank you for your patience in my indecision 😄.

BenjaminPrieur:
up

hishnash:
@daltoniam Im currently working on a branch that moves the connection logic out of the main WebSocket class into a dedicated class. The idea being users can switch in and out that connection class if needed to support things like custom proxies etc and taking over of existing connections. Maybe we could merge this concept of mocking into that in that there is a dedicated connection class that is set up for mocking that would not pollute the main code base.

daltoniam:
I did some refactor work for Swift 4 (version 3.0.0) and add this in. Sorry it took so long to get in, but all joy now 😄. Thanks for the PR!

PR #323, opened 2017-03-27
Avoid a warning caused by len < 0 where len is an UInt64

Timac:
When using Xcode 8.3b5, you will see a warning due to the comparison len < 0 where len is an UInt64.


dfed:
Would love to see this merged. Was about to create a similar PR.

daltoniam:
Thank you for this, but I resolved this when I did a SSL pinning fix. `2.0.4` is now released! Thanks again for taking the time to make a PR.

PR #324, opened 2017-03-27
Avoid a warning when the receivedOpcode raw value  is unknown

Timac:
Avoid a warning when the receivedOpcode raw value  is unknown with Xcode 8.3b5


nuclearace:
Your other PR is enough I believe.

daltoniam:
I agree with @nuclearace. `2.0.4` includes this warning fix. Thanks!

PR #326, opened 2017-03-29
Added parameters to configure queue priorities for the WebSocket (revised)

Elethier:
#304 was getting rather stale, and I finally found time to fix up the issues brought up in code review. Here's a revised pull request. I added a read and write queue with configurable queue priorities, and the write operation queue uses the write queue as its underlying queue.

I tried to remove the sharedWorkQueue, but I wasn't able to find a clean way to do it. Since disconnect code can be called on the read _or_ write queue, I needed a third queue to consolidate the operations to a single queue. I'm open to any suggestions. 

I was able to successfully run the Autobahn tests, and the code passes all of the implemented tests. I made sure to run the tests multiple times, to make sure I didn't introduce any weird race conditions as a result of the queue changes.

@daltoniam please review.


jakeholland:
@daltoniam I will be taking over this PR for @Elethier. Do you have any other questions or comments?

Source/WebSocket.swift
```diff
@@ -860,8 +879,9 @@ open class WebSocket : NSObject, StreamDelegate {
     /**
      Used to write things to the stream
      */
-    private func dequeueWrite(_ data: Data, code: OpCode, writeCompletion: (() -> ())? = nil) {
+    private func dequeueWrite(_ data: Data, code: OpCode, priority: Operation.QueuePriority = .normal, writeCompletion: (() -> ())? = nil) {
```
hishnash: With respect to these queue does this mean that if I send a low priority then a high priority the high priority would be sent first or is the order maintained?

jakeholland: Order is not maintained across priority in an OperationQueue so if you sent a low priority and then a high priority it may go ahead of the low priority.

acmacalister:
The codebase has actually under gone a major refactor, so I'm going to close this for now. Feel free to check it out here #653. Feedback welcomed!

PR #331, opened 2017-05-04
Added docs and error message for SSL Cipher Suites

### LINKED ISSUE #321: Usage of enabledSSLCipherSuites is unclear (to me)
I'm trying to connect to a HTTPS-based Websocket server, using 

    let socket = WebSocket(url: URL(string: "wss://example.com/etc/etc/")!)
    socket.connect()

This fails however, in Websocket.swift:341

    if let cipherSuites = self.enabledSSLCipherSuites

Here, `self.enabledSSLCipherSuites` is `nil`, so execution returns without an error message. Is this the intended behaviour? If not, I could create a PR that prints out some kind of error message or throws an exception.

I can't find relevant docs for how to use `enabledSSLCipherSuites`, I haven't seen it in the example code, and the property is not used anywhere else that I can see. Could you maybe give a short example on how to use this, or how I can make my connection work?

winsmith:
When trying to connect to a HTTPS-based web server, users will have to specify supported SSL cipher suites. This PR includes documentation that points out this fact, as well as an error message for users who have not set the `enabledSSLCipherSuites` property.

This fixes #321


Source/WebSocket.swift
```diff
@@ -355,6 +355,9 @@ open class WebSocket : NSObject, StreamDelegate {
                         return
                     }
                 }
+            } else {
```
daltoniam: I believe adding this else branch will cause anyone not using SSL ciphers on an encrypted socket to fall into this, no? I think we would only want to fall into this if SSL ciphers are added? It isn't necessarily required to use SSL ciphers to connect to an encrypted socketed. 

P.S. I really like the descriptive error message. Nice work.

winsmith: Hi, thanks for the nice response. I'm a bit confused, because I didn't realise that it's possible to connect to an encrypted socket without ciphers (it doesn't work on the server I'm testing on). Do you have a suggestion what condition to test for, so that the error message is only shown when the connection actually fails?

daltoniam: hmmm, I'm not sure how to easily keep the error message. The error would probably make more sense if we could send that error from a stream event that said SSL failed to connect because of the missing ciphers, but that seems like it would have to come from the system :confusing.

winsmith: OK, how about I only submit the docs for now, and I'll try to think of how to introduce the error message in a separate later pull request?

daltoniam:
Thanks!

PR #333, opened 2017-05-21
Remove void parameter from onConnect, Swift4 would require passi…

nuclearace:
…ng () into it

@daltoniam Noticed in the Swift 4 preview that this would require passing `()` into the onConnect callback, which is a bit funky.


nuclearace:
This is is backwards compatible.

daltoniam:
Agreed. I haven't look at an actual Swift 4 snapshot yet, probably should soonish 😄. Thanks!

PR #337, opened 2017-05-24
Add Advanced Delegate and Custom HTTP methods

hishnash:
Added the `WebSocketAdvancedDelegate` protocol that optionaly replaces the `WebSocketDelegate` adding support for:

* More detail when getting data (exposes the WSFrame object to the delegate)
* More detail about the HTTP request and response so that the user can extract needed information.

Also added support for custom HTTP methods for the http request.


acmacalister:
I dig this. I like the idea of merging this with #298 like you suggested. Going to run it against the AutoBahn Test Suite tomorrow to make sure everything is kosher, then I will merge.

PR #339, opened 2017-05-25
Support websocket compression

### LINKED ISSUE #169: Compression Support
Is compression something you plan to support in the future?

https://tools.ietf.org/html/draft-ietf-hybi-permessage-compression-28

joejune:
Fixes #169.

Using the Autobahn test suite, I confirmed that these changes perform comparably with the Autobahn library itself.  See attached report:

[autobahn_test.html.zip](https://github.com/daltoniam/Starscream/files/1027602/autobahn_test.html.zip)

### DIFF
--- a/Source/Compression.swift
+++ b/Source/Compression.swift
@@ -41,14 +41,14 @@ class Decompressor {
         guard initInflate() else { throw NSError() }
     }
     
-    func decompress(_ data: Data) throws -> Data {
+    func decompress(_ data: Data, finish: Bool) throws -> Data {
         let data = data
         let tail = Data([0x00, 0x00, 0xFF, 0xFF])
         
         var decompressed = Data()
         
         try decompress(in: data, out: &decompressed)
-        try decompress(in: tail, out: &decompressed)
+        if finish { try decompress(in: tail, out: &decompressed) }
         
         return decompressed
         
@@ -68,8 +68,7 @@ class Decompressor {
                 
                 let byteCount = buffer.count - Int(strm.avail_out)
                 out.append(buffer, count: byteCount)
-            }
-                while res == Z_OK && strm.avail_out == 0
+            } while res == Z_OK && strm.avail_out == 0
             
         }
         guard (res == Z_OK && strm.avail_out > 0)
--- a/Source/Compression.swift
+++ b/Source/Compression.swift
@@ -1,10 +1,30 @@
+//////////////////////////////////////////////////////////////////////////////////////////////////
 //
 //  Compression.swift
-//  Starscream
 //
-//  Created by Joseph Ross on 5/23/17.
-//  Copyright © 2017 Vluxe. All rights reserved.
+//  Created by Joseph Ross on 7/16/14.
+//  Copyright © 2017 Joseph Ross.
 //
+//  Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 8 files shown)


Source/WebSocket.swift
```diff
@@ -577,6 +597,34 @@ open class WebSocket : NSObject, StreamDelegate {
         }
         if let cfHeaders = CFHTTPMessageCopyAllHeaderFields(response) {
             let headers = cfHeaders.takeRetainedValue() as NSDictionary
+            if let extensionHeader = headers[headerWSExtensionName as NSString] as? NSString {
+                let parts = extensionHeader.components(separatedBy: ";")
```
hishnash: for readability what about moving this out to another method. There are already some uber methods on this class personally my preference on this is to keep method size down where it makes logical sense.

Source/Compression.swift
```diff
@@ -0,0 +1,201 @@
+//
+//  Compression.swift
+//  Starscream
+//
+//  Created by Joseph Ross on 5/23/17.
+//  Copyright © 2017 Vluxe. All rights reserved.
+//
+
```
hishnash: Link to the spec here so that later if there is a bug one can quickly find the reference to follow: https://tools.ietf.org/html/rfc7692

hishnash:
Consider adding a short section to the readme for users.

joejune:
@hishnash: Thanks for the great feedback.  I've made all the changes you suggested.

rakeshpethani:
When is this support going to be available ?

y0sh33:
+1 Also interested in this

joejune:
@daltoniam: Is there anything I can do to help this PR land?  Would it be better if I made compression disabled by default, and those who want to use it have to opt-in?

daltoniam:
@joejune thank you much for this, it looks great. Also sorry for the delay, got distracted with life and WWDC. Love the tests, README updates, autobahn tests, very through much appreciated.

nuclearace:
Did you test this with the Swift Package Manager? I have a feeling it might not like that you've included zlib outside of Sources.

daltoniam:
ah right, package managers 😄. I have not done that yet, might have to fix that.

daltoniam:
@nuclearace is there a way to test the package manager without have to create a new tag?

nuclearace:
Uh, if you create an app that has Starscream as a dependency and then put Starscream into edit mode.

nuclearace:
Actually I take that back, that wouldn't really help test if the dependency itself is right. I can't remember if they've added the ability to point to a branch

daltoniam:
Ok cool. Yeah, think it is going to be in Swift 4. 

https://github.com/apple/swift-evolution/blob/master/proposals/0150-package-manager-branch-support.md

daltoniam:
I guess I'll just create a test tag and work off of that (`2.1.0`).

daltoniam:
Found something I think might help:
https://github.com/apple/swift-package-manager/blob/master/Documentation/Usage.md#require-system-libraries

daltoniam:
@nuclearace I figured it out. I updated the 2.1.0 to build properly again.

joejune:
Thanks for getting this merged in.  I haven't worked with Swift Package Manager yet, so thanks for taking care of that aspect.

Issue #358, opened 2017-07-14
When processing a HTTP header, there is an off-by-one bug that results in parsing error.


dinghaoz:
if the header is 
"HTTP/1.1 440 Client Error (440)\r\n\r\n{"api:statuscode": 105, "api:message": "Invalid session. Please re-login."}" 
the `validateResponse()` returns 200, but 440 expected

```
for i in 0..<bufferLen {
    if buffer[i] == CRLFBytes[k] {
        k += 1
        if k == 3 {
            totalSize = i + 1
            break
        }
    } else {
        k = 0
    }
}
```

` if k == 3` should be ` if k == 4`

daltoniam:
See #359.

PR #359, opened 2017-07-14
off-by-one bug

dinghaoz:
if the header is
"HTTP/1.1 440 Client Error (440)\r\n\r\n{"api:statuscode": 105, "api:message": "Invalid session. Please re-login."}"
the validateResponse() returns 200, but 440 expected

### DIFF
--- a/Source/WebSocket.swift
+++ b/Source/WebSocket.swift
@@ -595,7 +595,7 @@ open class WebSocket : NSObject, StreamDelegate {
         for i in 0..<bufferLen {
             if buffer[i] == CRLFBytes[k] {
                 k += 1
-                if k == 3 {
+                if k == 4 {
                     totalSize = i + 1
                     break
                 }


daltoniam:
Can you provide any more information on this? I saw #358 but I'm not exactly sure what you mean? The status code always returns a 200 without this change?

dinghaoz:
If you receive a http response with header like

"HTTP/1.1 440 Client Error (440)\r\n\r\n{"api:statuscode": 105, "api:message": "Invalid session. Please re-login."}"

validateResponse() should returns 440 instead of 200.

daltoniam:
I tried this change out and it just breaks the ability to connect to any WebSocket server. Also the validate response coming back wrong doesn't make sense, since the code doesn't have any hardcode responses. I'm gonna guess that there is a specific issue in your environment, but thanks for taking the time to make a PR.

PR #364, opened 2017-07-28
Namespace zlib and common crypto

### LINKED ISSUE #361: Redefinition of module 'CommonCrypto'
Hi,

I'm using your framework in combination with several others - including [CommonCrypto](https://github.com/soffes/CommonCrypto). Unfortunately both frameworks - Starscream and CommonCrypto - are using the **system** CommonCrypto framework.

This results in a duplication error in the modulemaps:
`
.../Starscream/zlib/module.modulemap:6:8: Redefinition of module 'CommonCrypto'
...
.../CommonCrypto.framework/Modules/module.modulemap:1:8: note: previously defined here
module CommonCrypto [system] {
       ^
`

I'm not sure what the appropriate fix for this problem is - #353 seems to address a similar issue. Renaming the module name `CommonCrypto` will probably work, but is this a clean solution?

Here's a link to the [CommonCrypto's modulemaps](https://github.com/soffes/CommonCrypto/tree/master/CommonCrypto) - I searched for a related issue there, but found nothing.

In a carthage-only scenario I would remove Starscream' s dependency of the CommonCrypto (System) framework and would replace it with an appropriate entry for the cartfile - but since your framework should be usable outside carthage this is not an ideal solution.

Any ideas how to fix this - for all environments?

I posted the issue as well in [CommonCrypto's issue section](https://github.com/soffes/CommonCrypto/issues/6).

nuclearace:
@daltoniam Take a look. This will fix #361. Needs the changes in the other repos for the SPM.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -25,7 +25,7 @@ let package = Package(
     dependencies: [
         .Package(url: "https://github.com/daltoniam/zlib-spm.git",
                  majorVersion: 1),
-        .Package(url: "https://github.com/IBM-Swift/CommonCrypto.git",
+        .Package(url: "https://github.com/daltoniam/common-crypto-spm",
                  majorVersion: 0),
         ],
     exclude: ["Tests", "examples"]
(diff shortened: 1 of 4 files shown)


nuclearace:
https://github.com/daltoniam/Starscream/pull/364/files?w=1 for a better diff

daltoniam:
Thanks!

PR #377, opened 2017-09-05
Reduce to two targets and a single scheme

hamchapman:
iOS, macOS, and tvOS can all be supported using a single scheme and two targets: one for the framework itself, one for the tests).

That's what this does. No functionality changes.

Note that the base branch is the `swift4` branch.


daltoniam:
Nice! Thanks.

PR #388, opened 2017-09-26
HTTP headers are case insensitive

nuclearace:
@daltoniam HTTP headers are case insensitive, so servers can send any case in the upgrade resp.


daltoniam:
👍

PR #390, opened 2017-09-27
Add macOS Deployment Target.

xjbeta:
Fix "Module file's minimum deployment target is OS X v10.13".


daltoniam:
Ah, can't believe I missed that. Thanks!

PR #396, opened 2017-10-05
infinite loop

wowlocal:
Let len can return 0, it happens when outpusStream is nil. This loop ```while !sOperation.isCancelled)```  will never stop because it will always fall into the else block ```total += len``` where total = 0 and len = 0

### DIFF
--- a/Sources/WebSocket.swift
+++ b/Sources/WebSocket.swift
@@ -1187,7 +1187,7 @@ open class WebSocket : NSObject, StreamDelegate, WebSocketClient, WSStreamDelega
                 let stream = s.stream
                 let writeBuffer = UnsafeRawPointer(frame!.bytes+total).assumingMemoryBound(to: UInt8.self)
                 let len = stream.write(data: Data(bytes: writeBuffer, count: offset-total))
-                if len < 0 {
+                if len <= 0 {
                     var error: Error?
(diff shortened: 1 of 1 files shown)


Sources/WebSocket.swift
```diff
@@ -479,7 +479,7 @@ open class WebSocket : NSObject, StreamDelegate, WebSocketClient, WSStreamDelega
      - parameter completion: The (optional) completion handler.
      */
     open func write(string: String, completion: (() -> ())? = nil) {
-        guard isConnected else { return }
+        guard !isConnected else { return }
```
nuclearace: This shouldn't change right?

wowlocal:
Right

Sources/WebSocket.swift
olesDem: This is not the root of the problem. It's 

 public func write(data: Data) -> Int {
        guard let outStream = outputStream else {return 0}
        let buffer = UnsafeRawPointer((data as NSData).bytes).assumingMemoryBound(to: UInt8.self)
        return outStream.write(buffer, maxLength: data.count)
    }

Problem

daltoniam:
I manually addressed this in version 3.0.3 (just released), but thank you for the PR.

PR #400, opened 2017-10-09
Fixes #399 String.characters.count deprecation warning in Xcode 9.1

### LINKED ISSUE #399: String characters count deprecation
When I integrate Starscream in my lib and run `pod lib lint` I get deprecation warnings:

```
- WARN  | xcodebuild:  Starscream/Sources/WebSocket.swift:534:35: warning: 'characters' is deprecated: Please use String or Substring directly
- WARN  | xcodebuild:  Starscream/Sources/WebSocket.swift:810:26: warning: 'characters' is deprecated: Please use String or Substring directly
- WARN  | xcodebuild:  Starscream/Sources/WebSocket.swift:811:33: warning: 'characters' is deprecated: Please use String or Substring directly
```

fassko:
String.characters.count deprecation warnings

### DIFF
--- a/Sources/WebSocket.swift
+++ b/Sources/WebSocket.swift
@@ -531,7 +531,7 @@ open class WebSocket : NSObject, StreamDelegate, WebSocketClient, WSStreamDelega
         request.setValue("\(url.host!):\(port!)", forHTTPHeaderField: headerWSHostName)
 
         var path = url.absoluteString
-        let offset = (url.scheme?.characters.count ?? 2) + 3
+        let offset = (url.scheme?.count ?? 2) + 3
         path = String(path[path.index(path.startIndex, offsetBy: offset)..<path.endIndex])
         if let range = path.range(of: "/") {
             path = String(path[range.lowerBound..<path.endIndex])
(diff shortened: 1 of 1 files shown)


Sources/WebSocket.swift
```diff
@@ -531,7 +531,7 @@ open class WebSocket : NSObject, StreamDelegate, WebSocketClient, WSStreamDelega
         request.setValue("\(url.host!):\(port!)", forHTTPHeaderField: headerWSHostName)
 
         var path = url.absoluteString
-        let offset = (url.scheme?.characters.count ?? 2) + 3
+        let offset = (url.scheme?.count ?? 2) + 3
```
dfed: This change will break for anyone not yet on Xcode 9. Might be best to do the following instead:
```
#if swift(>=3.2)
let offset = (url.scheme?.count ?? 2) + 3
#else
let offset = (url.scheme?.characters.count ?? 2) + 3
#endif
```

dfed: Nevermind just saw we're already Swift 4 🙂

fassko: This deprecation message actually is only in Xcode 9.1 but let's get ready for it. ;)

daltoniam:
Thanks! Let us get ahead of the curve for once 😄.

PR #404, opened 2017-10-13
Fix version number in changelog

rastersize:
It said 3.0.1 when it should be 3.0.2 😸


daltoniam:
tsk, tsk, my bad copy and paste work 😄. Thanks!

PR #412, opened 2017-10-20
improved handling for self-signed certificate validation

eliburke:
separate disableCertValidation from overrideTrustHostname
add a 2nd method of getting the remote host domain name in sslTrust()


eliburke:
@daltoniam hey can you let me know if you have an opinion or any complaints on these changes? I'd be happy to explain all of them in more detail, or make changes if you prefer.

daltoniam:
@eliburke took me longer than I would like, but finally got around to taking a look at this. The changes make sense, I like it. Nice work, thanks!

PR #418, opened 2017-11-07
Fix warning for Swift 4 on the use of `character` property

frederik-jacques:
Hi,

This PR fixes a warning on Xcode 9 / Swift 4 for using the `characters` property.


daltoniam:
Ah good catch. We already removed the `characters` in #400, but I missed the new one from #412. Thanks!

fassko:
@daltoniam can you release update with these things to silence the warning? As the Xcode 9.1 is out.

PR #420, opened 2017-11-07
Fix: tvOS framework deployment target

### LINKED ISSUE #419: tvOS deployment_target inconsistencies
The podspec states that the `deployment_target` for tvOS is 9.0, while the build settings for the framework specify it should be 10.0.

This means that installing through CocoaPods would allow the framework to be used against tvOS 9.0, while installing it with Carthage would require 10.0.

The tests ran without errors when I changed the build settings for 9.0, should I submit a PR with this change?

fjcaetano:
The podspec states that the deployment_target for tvOS is 9.0, while the build settings for the framework specify it should be 10.0.

This means that installing through CocoaPods would allow the framework to be used against tvOS 9.0, while installing it with Carthage would require 10.0.

The tests ran without errors on tvOS 9.0 simulator.

fix #419


daltoniam:
Good catch, thanks!

Issue #240, opened 2016-08-25
Using with System Proxy on Simulator


kevinmlong:
It seems that when using this framework on a simulator (I've been using it for a while now, but never on the simulator), the connection request ignores the system proxy. Other network requests (e.g. REST API calls using Alamofire) are handled appropriately and routed through the proxy. IS there a way to set the proxy for this framework or have others experienced this issue?

acmacalister:
I haven't heard of any issues with proxies. I would assume since proxies are suppose to be transparent and we are relying the Apple networking APIs just like AlamoFire does, it should work fine.

kevinmlong:
So, after hours of exploring, researching, and testing, I've got it working with the following:

```swift
//WebSocket.swift
//This was already here
inputStream = readStream!.takeRetainedValue()
outputStream = writeStream!.takeRetainedValue()
//Added code below        
let proxyDict = CFNetworkCopySystemProxySettings()
let socksConfig = CFDictionaryCreateMutableCopy(nil, 0, proxyDict!.takeRetainedValue());
        
CFWriteStreamSetProperty(outputStream, kCFStreamPropertySOCKSProxy, socksConfig);
CFReadStreamSetProperty(inputStream, kCFStreamPropertySOCKSProxy, socksConfig);
```
This picks up the system wide proxy settings. Additionally, it seems that it only observes the SOCKS proxy setting. I've been able to get it working by running Charles Proxy on the host machine with External Proxy settings only set for HTTP/HTTPS.

daltoniam:
Ah great work and thanks for taking the time to figure this out. Seems like the system proxy settings should be already be applied to the NSStream, but Cocoa APIs 😄. I'll make sure to look at this when I do the Swift 3 stuff. Thanks again.

ergunkocak:
Is this implemented?

skerkewitz:
The above code may run on the simulator (I didn't check TBH) but will not work on a real iOS device. The reason is that you can not specify a SOCK5 server in the iOS setting. Therefore `CFNetworkCopySystemProxySettings()` will return a dictionary that only contains HTTP proxy information.

Unfortunately the HTTP proxy is completely useless in this case as it will be ignore (you would have to create the socket using `CFReadStreamCreateForHTTPRequest()` (which is deprecated by the ways) as far as I understand it.

I think the easiest and simplest solution for everybody would be if you could set the sock5 proxy information on the WebSocket instand yourself before you call `connect()`. Somewhat like this:

```
[...]

PR #423, opened 2017-11-15
enableSOCKSProxy - See websocket traffic in Charles proxy

neoneye:
Here is a possible solution for #240 

# Usage

```
import Starscream

class AppSocket {
	private let url: URL

	private lazy var socket: WebSocket = {
		var request = URLRequest(url: self.url)
		request.timeoutInterval = 5
		let stream = FoundationStream()
		stream.enableSOCKSProxy = true
		let instance = WebSocket(request: request, protocols: ["coldwar_mad"], stream: stream)
		instance.delegate = self
		return instance
	}()

	init(url: URL) {
		self.url = url
	}

	func connect() {
		socket.connect()
	}
}

extension AppSocket: WebSocketDelegate {
	// ...
}

```


daltoniam:
Looks good! Thanks.

Issue #442, opened 2017-12-14
'count' is unavailable: there is no universally good answer


BlakeBarrett:
I updated to the latest version (3.0.3) and am now getting a strange error in four places.

My environment is:   
   • Xcode 8.3.3
   • Swift 3.2
   • Starscream 3.0.3

The **errors** I'm seeing are:  
* /…/[WebSocket.swift:270:86](https://github.com/daltoniam/Starscream/blob/master/Sources/WebSocket.swift#L270): 'count' is unavailable: there is no universally good answer, see the documentation comment for discussion
 * /…/[WebSocket.swift:593:35](https://github.com/daltoniam/Starscream/blob/master/Sources/WebSocket.swift#L593): 'count' is unavailable: there is no universally good answer, see the documentation comment for discussion
 * /…/[WebSocket.swift:877:26](https://github.com/daltoniam/Starscream/blob/master/Sources/WebSocket.swift#L877): 'count' is unavailable: there is no universally good answer, see the documentation comment for discussion
 * /…/[WebSocket.swift:878:33](https://github.com/daltoniam/Starscream/blob/master/Sources/WebSocket.swift#L878): 'count' is unavailable: there is no universally good answer, see the documentation comment for discussion

I imagine this would be simple enough to fix with:
```Swift
fileprivate extension String {
   var count: Int {
      return self.characters.count
   }
}
```

BlakeBarrett:
I am running Xcode 8.3.3 in Swift 3.2 language mode.
I assume this came about because the rest of the civilized world has upgraded to Xcode 9/Swift 4 by now.

BlakeBarrett:
[This Pull Request](https://github.com/daltoniam/Starscream/pull/443) fixes this issue (for me).

introrse:
I had the same problem using Xcode 8.3.3, also in Swift 3.2 mode, and a brand-new Carthage-based install.  I manually applied Blake's patch and it resolved the issue.

introrse:
Also, for anyone else wondering, Xcode 9.2 seems to work out of the box, without the patch mentioned above.  Which makes sense since Swift 4 made Strings collections again.

daltoniam:
ha, yeah I didn't do anything with Swift 3.2 😄. I pulled in your PR though (#443) into the latest 3.0.4 release. Thanks again.

Saifalip:
could help me i am also facing the same error 'count' is unavailable: there is no universally good answer.
i am using xcode 8.1 to run

[...]

PR #443, opened 2017-12-14
Fixes: https://github.com/daltoniam/Starscream/issues/442

BlakeBarrett:
Should build again under Swift 3.2


daltoniam:
Thanks! I had another branch going with some other fixes so I just pulled this code into version `3.0.4` which I just released! Thanks again.

PR #444, opened 2017-12-15
Fix invalid NSError instance

werm098:
Calling default init on NSError now throws an exception (Xcode 9.2)
Warning Message: `-[NSError init] called; this results in an invalid NSError instance. It will raise an exception in a future release. Please call errorWithDomain:code:userInfo: or initWithDomain:code:userInfo:. This message shown only once.`

### DIFF
--- a/Sources/Compression.swift
+++ b/Sources/Compression.swift
@@ -52,7 +52,7 @@ class Decompressor {
 
     func reset() throws {
         teardownInflate()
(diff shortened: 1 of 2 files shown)


daltoniam:
Ah good catch. I had another branch going with some other fixes so I just pulled this code into version 3.0.4 which I just released! Thanks again.

PR #448, opened 2017-12-25
Check if self is nil in infinite while loop. 

mbesnili:
Move duplicated errorWithDetail method to ErrorHandle struct.

Our ios application uses socket in a way that it closes the connection when the application will resign active and establish the new connection when application did become active. That means we're extensively close and open connection while monkey testing (pressing home button causes to resign active and become active). This patch creates a short circuit to infinite while loop if web socket stream was nil-ed. More importantly, this prevents new socket connection to being connected somehow.

### DIFF
--- a/Sources/WebSocket.swift
+++ b/Sources/WebSocket.swift
@@ -48,6 +48,7 @@ enum InternalErrorCode: UInt16 {
     case compressionError = 2
     case invalidSSLError = 3
     case writeTimeoutError = 4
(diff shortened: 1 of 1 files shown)


mbesnili:
@daltoniam could you please review this pr ?

daltoniam:
Thanks! I had another branch going with some other fixes so I just pulled this code into version 3.0.4 which I just released! Thanks again.

Issue #172, opened 2016-02-17
Unconditionally adding port to Host header


lucacorti:
At https://github.com/daltoniam/Starscream/blob/master/Source/WebSocket.swift#L226 the port number is unconditionally appended to the hostname in the `Host` header. Is there any good reason for doing that?

This seems redundant and breaks at least my load balancer.

daltoniam:
According to [RFC 2616](https://tools.ietf.org/html/rfc2616#section-3.2.2) it will assume port 80 which isn't ideal in all cases. I could see the issue with a load balancer though, so we could change it to where it doesn't append the port if it is in fact port 80. In the mean time, you can just override the header if you want:

https://github.com/daltoniam/Starscream#custom-headers

lucacorti:
Yes, that would be great. I'd also omit the port if it's 443.

PR #452, opened 2018-01-11
Enable editing Host header value

victorg1991:
I have the same problem that https://github.com/daltoniam/Starscream/issues/172, but the solution that you proposed is not valid because the host header is always overriden the custom one (if included).

This PR adds the ability to customize the value of the Host header

Any comment would be appreciated :D

Thanks!


daltoniam:
Fair enough. That is a bit different now with the latest version, but this PR makes sense. Thanks!

victorg1991:
Thanks!!

Issue #346, opened 2017-06-13
Linux support (WebSocket client)


pixelspark:
Would really like to have Starscream on Linux as WebSocket client (a WebSocket server is already available through Kitura-WebSocket but a WebSocket client is currently missing, and the server WG has not yet produced anything usable at this point as far as I know).

Would it be feasible to switch over to a generic socket library (such as https://github.com/IBM-Swift/BlueSocket ? I believe there is also SSL support in there for both macOS and Linux).

#285 discussed Linux support before, but that concerned a WebSocket _server_.

daltoniam:
I would like to see Linux support too 😄. I think it might be feasible, I know in #335 there has been talk of allowing a swap out of different socket connections/adapters which would probably be the most flexible route. It is probably worth looking at with the Xcode 9 and Swift 4 update that will be need here in the near future.

daltoniam:
I did some work with the 3.0.0 release to allow this. There is a new `WSStream` class that can be implemented to allow a custom socket to be used. I also add some os checks for linux around the security framework since that doesn't exist on non apple platforms. I haven't done a full linux test yet, but I'm hopeful it should be feasible now.

pixelspark:
@daltoniam cool stuff! Will have a look at it later on. How is/will security processing (e.g. the hashing required to establish a connection) (be) done on Linux? I suppose you need to use CommonCrypto on OSX/iOS and openssl on Linux (perhaps just use a wrapper library such as IBM-Swift/Cryptor, which has worked well for me)?

acmacalister:
@pixelspark, yeah that is what we are probably looking at. We haven't heard or seen any signs of the `Security.framework` making it's way over to Linux, so a wrapper around OpenSSL is probably going to be our best bet. As far as the sha1 hashing is concerned, we could use the OpenSSL version or implement our own in pure Swift like CryptoSwift does. For the certificate/trust validation, I think OpenSSL is going to be the only way. Good catch on sha1 hashing still using CC. Should be fairly easy to move that out of the `WSStream` class and make something flexible for all the platforms.

caiofbpa:
Any way I can help this move forward? I got IBM-Swift/Cryptor included in Starscream locally, but I couldn't manage to run Autobahn|Testsuite to check if it's all good...

wbarksdale:
@caiofbpa I would be interested in testing out this branch!

caiofbpa:
[...]

PR #454, opened 2018-01-18
Linux port using IBM's libraries

caiofbpa:
Related to #346 

This should work on macOS, iOS and Linux. I couldn't manage to run Autobahn|Testsuite on it, though. Can someone help? Also I'm not sure how to test on other platforms like tvOS and watchOS.

I'm not sure replacing SSCommonCrypto (like I did) is the best approach. Would you rather have a `#if os(Linux)` around the Cryptor usage, keeping the old code on the `#else`? Or maybe go for something a little more fancy like what was done in `WSStream` to allow polymorphism? This polymorphic route is a bit bumpy because it could mean dynamic dependencies on `Package.json` and `Cartfile` or coming up with a plugin mechanism involving multiple repos.

I personally like the idea of just using IBM's cross-platform libraries "everywhere" as the base while the Server Work Group (or Apple itself) doesn't come up with an official solution. For that we would have to make sure IBM's BlueCryptor (and later on BlueSocket) work on all targeted platforms.

I tried to add BlueCryptor to Cartfile, but apparently BlueCryptor does not support Carthage. Is that an impediment to using it in Starscream?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 # Xcode
 .DS_Store
 build
+.build
 *.pbxuser
 !default.pbxuser
 *.mode1v3
--- a/Package.resolved
+++ b/Package.resolved
@@ -36,6 +36,15 @@
           "revision": "2eb3aff0fb57f92f5722fac5d6d20bf64669ca66",
           "version": "1.1.0"
         }
+      },
+      {
+        "package": "Socket",
+        "repositoryURL": "https://github.com/IBM-Swift/BlueSocket",
+        "state": {
+          "branch": null,
+          "revision": "0dacaaa9482871142cb8dc869cf4b80396450e16",
+          "version": "0.12.86"
+        }
       }
     ]
   },
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,0 +1,43 @@
+{
+  "object": {
+    "pins": [
+      {
+        "package": "CommonCrypto",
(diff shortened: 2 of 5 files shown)


wbarksdale:
This isn't really getting to linux support because `SecTrust` and `SSLSettings` are still used in the WSStream protocol, i'm unsure how to get around those...

caiofbpa:
@daltoniam I'm hitting a lot of roadblocks in this port. Can you help me? I'm stuck on an error about converting `DispatchQueue` to `dispatch_queue_t`.

caiofbpa:
I decided to take the IBM's route and implement `WSStream` with BlueSocket. The connection is established successfully, but I need more help testing it, I'm sure it's not working correctly, but I have no idea why. @wbarksdale can you help me? @daltoniam what do you think?

willrevl:
This looks promising, may have some time to test tomorrow!

wbarksdale:
this is looking promising! I'll take it for a spin tomorrow

caiofbpa:
Although I managed to get it to compile on Linux and the connection seems successful, it's unable to send or receive messages.

I solved my problem using the WebSocketClient on https://github.com/vapor/engine instead, and I don't intend on moving forward with this port, so I'm closing this PR.

If @wbarksdale or anyone else wants to continue this work, the code will be available at https://github.com/caiofbpa/starscream/tree/support-cryptor

PR #458, opened 2018-01-24
Fix #436 Submit the client certificate

### LINKED ISSUE #436: client certificate-based access
The WebSocket does not seem to solve the situation where server need a client certificate.
In WebSocket.swift, It just validates the server certificate, but does not validate the client certificate.

xingchen966:
This is a self-signed certificate to solve the problem, but not universal, I hope you can help improve it.


daltoniam:
Thanks! I think #481 should be the more universal thing we are looking for.

PR #476, opened 2018-03-01
Fix crash caused by cleanup() being called from multiple threads

flagoworld:
If InputStream/OutputStream .close() or *SetDispatchQueue is called
from two threads at the same time, it will cause a crash. This adds a
lock to prevent the cleanup() method from executing concurrently. Once
the first one is complete, inputStream and outputStream and nil and so
simultaneous attempts to cleanup() become noops.

Is it possible that locking inside cleanup() has some unintended side-effects? Take a look at let me know. I am not as familiar with the codebase...

### DIFF
--- a/Sources/WebSocket.swift
+++ b/Sources/WebSocket.swift
@@ -139,6 +139,8 @@ open class FoundationStream : NSObject, WSStream, StreamDelegate  {
     private var outputStream: OutputStream?
     public weak var delegate: WSStreamDelegate?
     let BUFFER_MAX = 4096
+    
+    private let cleanupMutex = NSLock()
 	
 	public var enableSOCKSProxy = false
     
@@ -248,6 +250,8 @@ open class FoundationStream : NSObject, WSStream, StreamDelegate  {
     }
     
     public func cleanup() {
+        cleanupMutex.lock()
+        
         if let stream = inputStream {
             stream.delegate = nil
             CFReadStreamSetDispatchQueue(stream, nil)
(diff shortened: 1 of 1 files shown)


hamchapman:
@fassko or @flagoworld, have either of you tried this with any success, or is it a hunch that this could / should fix it?

Sources/WebSocket.swift
```diff
@@ -248,6 +250,8 @@ open class FoundationStream : NSObject, WSStream, StreamDelegate  {
     }
     
     public func cleanup() {
+        cleanupMutex.lock()
```
shawnce: The look will need to be applied universally to protect access to the var protected here or it won't help much. I would also use an NSRecursiveLock unless you can fully understand the calling model that exists.

flagoworld:
@hamchapman
I have had no further crashes since making this change.

Sources/WebSocket.swift
flagoworld: I don't think recursive lock is necessary here, but I understand what you are saying about protecting all the variables that are used within the lock across the class. That said, so far in practice this has eliminated the crashes without issue. Provide a better implementation for Dalton so we can get a fix merged!

daltoniam:
I spent some time to cleanup the multiple NSLock objects and also help with this issue. I release new version (`3.0.5`) that should have a fix for this. Let me know if that fixes it.

daltoniam:
It's been a while and based on the thumbs up feedback, I'm gonna assume we fix this. Thanks again!

PR #507, opened 2018-04-24
added open keyword to SSLSecurity.isValid, added public to other methods

eliburke:
After connecting (or failing to connect), I need to be able to access the raw SSL certs. 

The correct-ist solution is probably some delegate methods, but making SSLSecurity.isValid() an open function is a much simpler solution-- I can override isValid() and directly access _trust_ and _domain_.

### DIFF
--- a/Sources/SSLSecurity.swift
+++ b/Sources/SSLSecurity.swift
@@ -198,7 +198,7 @@ open class SSLSecurity : SSLTrustValidator {
     
     - returns: a public key
     */
-    func extractPublicKey(_ data: Data) -> SecKey? {
+    public func extractPublicKey(_ data: Data) -> SecKey? {
         guard let cert = SecCertificateCreateWithData(nil, data as CFData) else { return nil }
         
         return extractPublicKey(cert, policy: SecPolicyCreateBasicX509())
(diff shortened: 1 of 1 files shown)


daltoniam:
Yeah fair enough. I'm gonna merge it for now and think about what might work better for the 4.0.0 release. I really want to put some effort into the security code for the 4.0.0 release. Thanks!

adamkaplan:
@daltoniam FWIW, we use TrustKit to do SSL Pinning on iOS. It does proper pinning (via SPKI Hashing) with fallback methods to support iOS 9 through 12, macOS, etc. You might consider adding a `subspec` for ssl pinning that adds TrustKit support instead of building this all yourself... or providing instructions in the README. Pinning is very very fragile and complex code to write. Especially if you intended for multi-platform support as the security APIs available vary widely (even between iOS versions). Plus, there is support for reporting endpoints which are used to discover malicious SSL attacks in realtime.

https://github.com/datatheorem/TrustKit

I integrate Starscream with TrustKit similar to this:
```
    class BlockTrustValidator: SSLTrustValidator {
        let validator: YFTLSTrustValidatorBlock
        
        init(withValidator validator: @escaping YFTLSTrustValidatorBlock) {
            self.validator = validator
        }
        
        func isValid(_ trust: SecTrust, domain: String?) -> Bool {
            return validator(trust, domain)
        }
    }
```
and
```
websocket.security = BlockTrustValidator { (trust: SecTrustRef, domain: String?) in
    return TSKTrustDecisionShouldBlockConnection != TSKPinningValidator.evaluateTrust(trust, forHostname: domain);
}
````

PR #519, opened 2018-06-08
Always wait for currently executing operations to finish

mgrant34:
I am seeing a number of crashes caused by a nil `outputStream` in sslTrust(), **release 3.0.5**.
 I can confirm that the issue is a nil output stream because the crashes are randomly occurring on lines 266, 267, and 269 in WebSocket.swift which are all attempting to access/force unwrap `outputStream`. Thus, the issue most likely is an async issue where `outputStream` is getting set to nil at different times.

This PR attempts to fix the issue by waiting until currently executing operations have completed before we set `outputStream` to nil, and avoid another concurrentOperation from accessing a nil `outputStream`.

**Per the documentation on `cancelAllOperations`:** https://developer.apple.com/documentation/foundation/nsoperationqueue/1417849-cancelalloperations 

"Canceling the operations does not automatically remove them from the queue or stop those that are currently executing. For operations that are queued and waiting execution, the queue must still attempt to execute the operation before recognizing that it is canceled and moving it to the finished state. For operations that are already executing, the operation object itself must check for cancellation and stop what it is doing so that it can move to the finished state. In both cases, a finished (or canceled) operation is still given a chance to execute its completion block before it is removed from the queue."

### DIFF
--- a/Sources/WebSocket.swift
+++ b/Sources/WebSocket.swift
@@ -710,11 +710,11 @@ open class WebSocket : NSObject, StreamDelegate, WebSocketClient, WSStreamDelega
      Disconnect the stream object and notifies the delegate.
      */
     private func disconnectStream(_ error: Error?, runDelegate: Bool = true) {
-        if error == nil {
-            writeQueue.waitUntilAllOperationsAreFinished()
-        } else {
+        if error != nil {
             writeQueue.cancelAllOperations()
         }
+
+        writeQueue.waitUntilAllOperationsAreFinished()
         
         mutex.lock()
         cleanupStream()


Sources/WebSocket.swift
```diff
@@ -710,11 +710,11 @@ open class WebSocket : NSObject, StreamDelegate, WebSocketClient, WSStreamDelega
      Disconnect the stream object and notifies the delegate.
      */
     private func disconnectStream(_ error: Error?, runDelegate: Bool = true) {
-        if error == nil {
-            writeQueue.waitUntilAllOperationsAreFinished()
-        } else {
+        if error != nil {
             writeQueue.cancelAllOperations()
```
mgrant34: cancelAllOperations DOES NOT stop currently executing blocks from finishing. We must wait until those have finished before we set `outputStream` to nil as the blocks are attempting to access `outputStream`.

Sources/WebSocket.swift
```diff
@@ -710,11 +710,11 @@ open class WebSocket : NSObject, StreamDelegate, WebSocketClient, WSStreamDelega
...
         }
+
+        writeQueue.waitUntilAllOperationsAreFinished()
```
mgrant34: We should always wait, as cancel does not stop currently executing operations, or those queued for execution.

adamkaplan:
@mgrant34 The alternative fix #524 was merged, meaning the specific crash we were all seeing will no longer happen (no more force unwrap).

However, this PR gets to the root of the issue and should still be discussed.

Sources/WebSocket.swift
adamkaplan: >While the current thread is blocked, the receiver continues to launch already queued operations and monitor those that are executing. During this time, the current thread cannot add operations to the queue, but other threads may. Once all of the pending operations are finished, this method returns.

It seems like it is possible for that call to block the calling thread indefinitely if the socket is not already in a failed/error state.  Another thread communicating over the socket would still allowed, and `ping` or other custom protocol features would still occur. Might need additional protection of the socket to prevent further use...

mgrant34: When there is no error, the behavior that this PR is proposing is the same exact behavior as previously. 

This PR essentially is only changing the behavior when there IS an error, per the diff.

acmacalister:
The codebase has actually under gone a major refactor and removed the whole public isConnected. Feel free to check it out here #653. Feedback welcomed!

PR #524, opened 2018-06-21
Fix crash nil outstream

### LINKED ISSUE #523: Crash in TLS
I'm seeing a crash that occurs 200+ times a day in Starscream. The stack traces are all identical and one is featured below. It is caused by the process of force-unwrapping an optional value: the SSL Peer/Host name from the output stream.

[Failing line of code](https://github.com/adamkaplan/Starscream/blob/master/Sources/WebSocket.swift#L267)

AppleDoc for [kCFStreamSSLPeerName](https://developer.apple.com/documentation/cfnetwork/kcfstreamsslpeername)

Notes:
- Potentially related to #517, which also mentions `CFReadStreamCopyProperty`
- I am doing SSL Pinning using custom `SSLTrustValidator` subclass bridging into the `TrustKit` pod

```
WebSocket.swift in FoundationStream.sslTrust() at line 267 (application)
WebSocket.swift in FoundationStream.sslTrust() at line 266 (application)
WebSocket.swift in protocol witness for WSStream.sslTrust() in conformance FoundationStream (application)
WebSocket.swift in closure #1 in closure #1 in WebSocket.initStreamsWithData(_:_:) at line 675 (application)
WebSocket.swift in thunk for @escaping @callee_guaranteed () -> () (application)
Foundation in __NSBLOCKOPERATION_IS_CALLING_OUT_TO_A_BLOCK__ + 12
Foundation in -[NSBlockOperation main] + 68
Foundation in -[__NSOperationInternal _start:] + 844
Foundation in __NSOQSchedule_f + 400
libdispatch.dylib in libdispatch.dylib + 6752
libdispatch.dylib in libdispatch.dylib + 40596
libdispatch.dylib in libdispatch.dylib + 34764
libdispatch.dylib in libdispatch.dylib + 60588
libdispatch.dylib in libdispatch.dylib + 59900
libsystem_pthread.dylib in _pthread_wqthread + 1172
libsystem_pthread.dylib in start_wqthread + 0
```

adamkaplan:
Expected to fix #523 

Looks like there is a code path in which websocket.sslTrust() may be called when outputStream is `nil`. The cast to create `domain` is unsafe, because `nil as! String` *may crash*, while `nil as! String?` will never crash.

Using an optional `String?` here is perfectly fine because a `String` would be cast to `String?` anyway when returned in the tuple `(SecTrust?, String?)`.

I have also added a guard at the top which itself will *probably* fix the issue. IMO having both safeguards in place is the best practice. The`outputStream` *should* not be nil for this function to produce anything useful (hence the guard). While it is technically still possible for `kCFStreamSSLPeerName` to be nil, although it should not be.

### DIFF
--- a/Sources/WebSocket.swift
+++ b/Sources/WebSocket.swift
@@ -263,8 +263,10 @@ open class FoundationStream : NSObject, WSStream, StreamDelegate  {
     #if os(Linux) || os(watchOS)
     #else
     public func sslTrust() -> (trust: SecTrust?, domain: String?) {
-        let trust = outputStream!.property(forKey: kCFStreamPropertySSLPeerTrust as Stream.PropertyKey) as! SecTrust?
-        var domain = outputStream!.property(forKey: kCFStreamSSLPeerName as Stream.PropertyKey) as? String
+        guard let outputStream = outputStream else { return (nil, nil) }
+
+        let trust = outputStream.property(forKey: kCFStreamPropertySSLPeerTrust as Stream.PropertyKey) as! SecTrust?
+        var domain = outputStream.property(forKey: kCFStreamSSLPeerName as Stream.PropertyKey) as! String?
         if domain == nil,
             let sslContextOut = CFWriteStreamCopyProperty(outputStream, CFStreamPropertyKey(rawValue: kCFStreamPropertySSLContext)) as! SSLContext? {
             var peerNameLen: Int = 0
--- a/Tests/CompressionTests.swift
+++ b/Tests/CompressionTests.swift
@@ -52,8 +52,9 @@ class CompressionTests: XCTestCase {
         // 2 Gigs!
 //        var rawData = Data(repeating: 0, count: 0x80000000)
         var rawData = Data(repeating: 0, count: 0x80000)
+        let rawDataLen = rawData.count
         rawData.withUnsafeMutableBytes { (ptr: UnsafeMutablePointer<UInt8>) -> Void in
-            arc4random_buf(ptr, rawData.count)
+            arc4random_buf(ptr, rawDataLen)
         }
         
         let compressed = try! compressor.compress(rawData)


Sources/WebSocket.swift
```diff
@@ -263,8 +263,10 @@ open class FoundationStream : NSObject, WSStream, StreamDelegate  {
     #if os(Linux) || os(watchOS)
     #else
     public func sslTrust() -> (trust: SecTrust?, domain: String?) {
-        let trust = outputStream!.property(forKey: kCFStreamPropertySSLPeerTrust as Stream.PropertyKey) as! SecTrust?
-        var domain = outputStream!.property(forKey: kCFStreamSSLPeerName as Stream.PropertyKey) as? String
+        guard let outputStream = outputStream else { return (nil, nil) }
+
+        let trust = outputStream.property(forKey: kCFStreamPropertySSLPeerTrust as Stream.PropertyKey) as! SecTrust?
+        var domain = outputStream.property(forKey: kCFStreamSSLPeerName as Stream.PropertyKey) as! String?
```
texuf: we usually follow the pattern `property as? String`, instead of the `as!` in the middle. I believe it gives the same result.

adamkaplan: You're suggesting the existing code... https://github.com/adamkaplan/Starscream/blob/master/Sources/WebSocket.swift#L267

I understand optional casts as a general best practice, however it isn't practical to use `as?` here for the reasons I explained in detail in the description above and the linked issue.

(it crashes! To now crash it would be`as? String?`, but then you end up with a `String??`...

adamkaplan:
Don’t worry about it! Took me a few moments to realize why that works too.

mgrant34:
We are also seeing a lot of crashes caused by nil output stream. This PR attempts to solve the underlying issue https://github.com/daltoniam/Starscream/pull/519

Either way, this PR or both should be merged in asap.

jrendel:
Is there a consensus on the proper/best fix for the sslTrust() crash (this PR or https://github.com/daltoniam/Starscream/pull/519?) We are seeing it quite frequently as well and would like to pull in the fix for our next release even if its not officially merged into the library yet.

adamkaplan:
@mgrant34 @jrendel My team deployed using my branch of StarScream late last week. Steady rate of ~500 crashes/day in StarScream has dropped to 0 in the updated version.

I suggest you deploy off my branch until @daltoniam comes back and merges this. Put this in your Podfile:
```
# Override Starscream until the fix: https://github.com/daltoniam/Starscream/pull/524
pod 'Starscream', :git => 'git@github.com:adamkaplan/Starscream.git', :branch => 'fix-crash-nil-outstream'
```

daltoniam:
Sorry about the delay. Looks good! Thanks!

adamkaplan:
thanks @daltoniam

vonox7:
@daltoniam could you please create a new release? Seems like a lot (or all) of the library users are affected by this crash. Also projects like https://github.com/socketio/socket.io-client-swift could then update to the new version.

mgrant34:
+1 I have also created an issue https://github.com/daltoniam/Starscream/issues/534 with 6 likes, even a minor version bump would really help a lot of us. Please!

Issue #526, opened 2018-06-25
Stream is not disconnecting when `disconnect` is not called with any arguments


brian-dlee:
Looking at the `disconnect(forceTimeout, closeCode)` method, it doesn't look like it ever hits `disconnectStream(nil)` unless you supply a timeout to allow the disconnect to be asynchronous.

You can actually get the `default` condition to be met if you pass a non-positive number to `forceTimeout`. I'm assuming this is a bug. Should `disconnectStream` be called immediately when no `forceTimeout` is provided?

brian-dlee:
If anyone else if being held up by this calling disconnect like this will work around the issue.

```swift
socket.disconnect(forceTimeout: -1, closeCode: CloseCode.normal.rawValue)
```

brian-dlee:
I forked and went to perform the change for this, but it turns out this behavior is intentional per comments in the code. It looks like the WebSocket server is not honoring the close code in my case.

daltoniam:
Addressed in PR above.

PR #527, opened 2018-06-25
Documents `disconnect` behavior in README

brian-dlee:
I was confused by some behavior in `disconnect` that is only documented within the code. I've just added this information to the README to help others.

See #526.


daltoniam:
Ok cool. Thanks!

PR #535, opened 2018-07-24
WIP: Add conditional support for Network.framework

davbeck:
I had some extra time so I thought I would give Network support a shot. It was actually surprisingly easy, although the SSL settings aren't implemented yet (hence the WIP). I tested it out using my own app that uses [socket.io](https://github.com/socketio/socket.io-client-swift) and it seems to be working without issue. It even feels a little faster, although that's probably just in my head 😅.

I figured this could be a good jumping off point.


daltoniam:
Nice work! Thanks for the PR. I started a [Xcode-10 branch here](https://github.com/daltoniam/Starscream/tree/xcode-10). I took the time to address a bit of refactoring I wanted to do too. That branch has a working implementation of the `Network.framework` as far as my tests have shown, but I'd look to get some feedback on it. Thanks again!

PR #554, opened 2018-09-21
Support custom workQueue in FoundationStream

### LINKED ISSUE #551: Multiple connection communication issue
When there are 2 connections and one of them is connecting, the other connection cannot send / receive any data.

I suspect it has something to do with the `sharedWorkQueue` here but I'm not sure...
https://github.com/daltoniam/Starscream/blob/70fd033db6d1d0300f900bd05fd05c1246eb721d/Sources/WebSocket.swift#L137

akiroz:
Resolves #551

### DIFF
--- a/Sources/Starscream/WebSocket.swift
+++ b/Sources/Starscream/WebSocket.swift
@@ -135,14 +135,20 @@ public protocol WSStream {
 }
 
 open class FoundationStream : NSObject, WSStream, StreamDelegate  {
-    private let workQueue = DispatchQueue(label: "com.vluxe.starscream.websocket", attributes: [])
+    private static let sharedWorkQueue = DispatchQueue(label: "com.vluxe.starscream.websocket", attributes: [])
+    private let workQueue: DispatchQueue
     private var inputStream: InputStream?
     private var outputStream: OutputStream?
     public weak var delegate: WSStreamDelegate?
     let BUFFER_MAX = 4096
 	
 	public var enableSOCKSProxy = false
-    
+   
+    public init(queue: DispatchQueue? = nil) {
+        workQueue = queue ?? FoundationStream.sharedWorkQueue
+        super.init()
+    }
+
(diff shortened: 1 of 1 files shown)


daltoniam:
Thanks for the PR! Based on the issue reported in #551, I don't think that is the direct source of the problem. I've used multiple sockets with the same work queue without issue, but there could be something specific to your use case. Do you happen to be pushing a lot of data on each socket (reading or writing)? I'm curious if maybe there is just a lot events getting added to the queue and thus causing issues. 

I'm not against this change, I just want to make sure we are making it for the right reasons. Thanks again!

akiroz:
@daltoniam It only happens when one of them is connecting, everything works after both are connected.

mikemm13:
@daltoniam I've just faced the same issue and I fixed in a similar way @akiroz did before seeing this thread. 
In my case, I try to connect 3 sockets at the same time and one of them is returning a timeout after 5 seconds. During that time, the rest of the sockets can't connect properly. 
On the disconnect callback from the failing socket, I have a retry mechanism that tries to connect after 5 seconds. Each time it tries to connect, is blocking the R/W operations from the other 2 sockets during the timeout period.
In my solution I create a sharedRWQueue that is the one used in CFReadStreamSetDispatchQueue and CFWriteStreamSetDispatchQueue

@daltoniam any chance you could check this again please?

fassko:
@akiroz could you please check the merge conflicts? Thanks!

akiroz:
@fassko Just rebased on master, all tests passing on Xcode 10.1

fassko:
@akiroz I think now there is code missing where this queue is going to be used.

akiroz:
Do you mean outside of this class with the same label as `com.vluxe.starscream.websocket`?

acmacalister:
The codebase has actually under gone a major refactor and completely changed how we handle our work queues. Feel free to check it out here #653. Feedback welcomed.

PR #520, opened 2018-06-14
Move CommonCrypto and zlib dependencies mapping to custom module map

turbulem:
Hi!
I faced an issue similar to [this](https://github.com/daltoniam/Starscream/issues/445) and this [this](https://github.com/daltoniam/Starscream/issues/437).
In my project I use Carthage and commit my artefacts into my repo. And CI(and any other machine) is not able to compile it without running 'carthage update'.
Basically the problem that to link Starscream.framework with other app or library you should provide module.map similar to the one, located in /zlib folder (but without header, which is required only for build).
To make this work I created new custom module map and added a wrapper for SHA1 algorithm used by Starscream. Also updated cocoapods spec to add custom module map + private headers.

Please consider this PR as a possible way of resolving issues linked above and also I'd be happy to hear from you any suggestions on this.

Thank you!
Cheers!

PR #482, opened 2018-03-10
Fix Redefinition of module error in Xcode ≧ 8.3, Take 2

vkedwardli:
Second attempt to fix this bug (first failed attempt:https://github.com/daltoniam/Starscream/pull/465)

Using dynamic generated modulemap now to avoid the Xcode bug, according to [Xcode 8.3 release notes](https://developer.apple.com/library/content/releasenotes/DeveloperTools/RN-Xcode/Chapters/Introduction.html):

> When building a framework that has module maps in both the source and the install directories that define the same module, the compiler will show a redefinition message. (28638816)

Reference:
https://github.com/ResearchKit/ResearchKit/issues/922
https://github.com/grpc/grpc/issues/10355

Should can fix the error mentioned by @rhishikeshj in  https://github.com/daltoniam/Starscream/issues/361#issuecomment-346810549

PR #556, opened 2018-09-23
update to swift 4 manifest

nuclearace:
The Swift 3 manifest will be going away in Swift 5.

@daltoniam I'm not sure the story about the PRs looking to fix the issues around the dependencies, but I'm thinking about moving those to internal system targets (something new introduced in Swift 4.2.) Thoughts?


daltoniam:
Thanks. I still needed to get this done, I appreciate you doing this. Yeah, #520 does fix an issue with Carthage builds, but it also introduces some Objective-C bridging which I don't want to do, because that would break Linux support (which I still want to do) and I think it might break the SPM builds. All these package managers never want to play nice together 😬. Thanks again.

nuclearace:
@daltoniam Yeah I actually tried to get linux working this weekend. It's going to be pretty painful I think, you'll have to get rid of the CFNetworking stuff. Since although the functions are _defined_ in open source core foundation, actually trying to use them runs into this gem.



So I think to really get linux support your best bet is to use SwiftNIO for that. But I'm not sure how that fits into the work with using the Networking lib. Since I don't think that is available on Linux either.

fassko:
Apple has released https://github.com/apple/swift-nio-transport-services for this task. Also there is https://github.com/apple/swift-nio-examples where some of examples will be kept.
In near future I'm looking more into SwiftNIO and how it can benefit also this project.

nuclearace:
@fassko Sweet. I think if this lib could use SwiftNIO for all platforms it would greatly ease the maintenance cost compared to a ton of `#if os` checks. (Which are probably still going to be needed)

daltoniam:
@nuclearace Well that is disappointing but not unexpected 😞. I agree with y'all, we should probably look at SwiftNIO. I might be holding on to an outdated version of the project, but I try to balance simplicity and bringing in a lot of dependencies which has been has been a bit of a pain with all the package managers. I guess that is already the status quo, so probably just makes sense to unify on a system.

vkedwardli:
@daltoniam  if #520 won't be merged, please consider https://github.com/daltoniam/Starscream/pull/482, thanks

PR #559, opened 2018-09-28
Move CommonCrypto and zlib dependencies mapping to custom module map

turbulem:
Details:
https://github.com/daltoniam/Starscream/pull/520


fassko:
Can you look into why it doesn't build with Swift PM. I get this error:

```
Compression.swift:30:8: error: no such module 'Starscream.SSCZLib'
import Starscream.SSCZLib
```

turbulem:
Sure, let me check

turbulem:
Looks like proposed solution cannot be adopted by SPM. It doesn't seem to support explicit modules. I will investigate what can be done here.

nuclearace:
I would see for SPM if you could just use Swift 4.2's system targets. They look something like this https://github.com/nuclearace/SwiftDiscord/tree/vapor3/Sources/Sodium and https://github.com/nuclearace/SwiftDiscord/blob/vapor3/Package.swift#L43

turbulem:
@nuclearace Thank you, it looks very relevant.
Also I think that in iOS12/Mojave CommonCrypto has system bindings in swift so if target is the latest SDK I can get rid of Obj-C helper

fassko:
Maybe this is the option we can aim. You can still use older release for lover Swift version.

fassko:
@turbulem looks like you're right about CommonCrypto https://stackoverflow.com/a/50690346

fassko:
In which version of Swift are these libraries included? I think need to explicitly set Swift 4.2 version in Package.swift and podspec file. .swift-version is going to be deprecated and isn't used anymore?

Starscream.podspec
```diff
@@ -11,11 +11,10 @@ Pod::Spec.new do |s|
   s.osx.deployment_target = '10.10'
   s.tvos.deployment_target = '9.0'
   s.watchos.deployment_target = '2.0'
-  s.source_files = 'Sources/*.swift'
-  s.libraries    = 'z'
+  s.source_files = 'Sources/**/*.{h,m,swift}'
+  s.module_map = 'Sources/modulemap/Starscream.modulemap'
+  s.private_header_files = 'Sources/modulemap/**/*.h'
```
fassko: If we delete modulemap and header file then this is not needed anymore.

Test also with some sample project integrating this pod against https://websocket.org/echo.html

Package.swift
```diff
@@ -27,10 +27,7 @@ let package = Package(
         products: [
```
fassko: I think let's set Swift 4.1 here as well likewise in podspec.

turbulem:
@fassko I updated my solution. It only works with new SDK, which might not be the best option. However it solves the problem and seem to work fine with Pods/Carthage/SPM.

Issue #565, opened 2018-10-05
Expose `betterPathUpdateHandler` and `viabilityUpdateHandler`


ollitapa:
Would be nice to get these notifications of `betterPathUpdateHandler` and `viabilityUpdateHandler` through either the WebSocketDelegate or by exposing NWConnection somehow. One would need to get those callbacks in before this: https://github.com/daltoniam/Starscream/blob/d14d4af2e7e3e32f9ce32ec6817bcd26ff7dc45b/Sources/Starscream/NetworkStream.swift#L70

Or would it be better to use NWPathMonitor outside of Starscream to get the network state and to decide wether to reconnect or not?

Also the `.waiting/.preparing` states could have a callback in WebSocketDelegate.

What do you think?

fassko:
@ollitapa I think that is a good idea. Maybe you could help with a PR?

ollitapa:
Sure, I'll make a proposal.

ollitapa:
Here is a simple proposal. In this one, the framework does not try to do any additional work if better path is available, but leaves it to the user to handle migration.

PR #577, opened 2018-10-10
Expose path viability and better path updates

ollitapa:
Proposal for exposing better path updates and path viability.
I'm not sure if the framework should handle migration. Maybe it's better let the user to do the switching if opening an other web socket needs some additional non-generic work done before it's good.

See #565

Good info was provided in WWDC2018 https://developer.apple.com/videos/play/wwdc2018/715/

### DIFF
--- a/Sources/Starscream/NetworkStream.swift
+++ b/Sources/Starscream/NetworkStream.swift
@@ -58,15 +58,26 @@ open class NetworkStream: WSStream {
             switch newState {
             case .ready:
                 doConnect(nil)
+            case .waiting:
+                self?.delegate?.streamIsWaitingForConnectivity()
             case .cancelled:
                 doConnect(nil)
             case .failed(let error):
                 doConnect(error)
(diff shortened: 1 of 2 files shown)


Sources/Starscream/NetworkStream.swift
```diff
@@ -58,15 +58,26 @@ open class NetworkStream: WSStream {
...
                 break
             }
         }
+
+        conn.viabilityUpdateHandler = { [weak self] (isViable) in
+            self?.delegate?.streamPathViabilityUpdate(isViable: isViable)
+        }
+
+        conn.betterPathUpdateHandler = { [weak self] (isBetter) in
+            self?.delegate?.streamBetterPathUpdate(isBetter: false)
```
mooshee: Shouldn't this be `isBetter: isBetter`?

ollitapa: Fix it!

fassko:
Thanks @ollitapa for your contribution!

PR #582, opened 2018-10-17
Refactor connection status as an enum

lickel:
We noticed a hole where our socket was getting disconnected and not reconnecting. 
In tracing down code paths, we noticed that the `isConnecting` and `didDisconnect` flags aren't currently thread safe. Further, only the `isConnected` state is exposed to consumers.

This exposes the underlying status as an enum, and guards access behind the same `mutex` that `connected` previously used.


Sources/Starscream/WebSocket.swift
```diff
@@ -57,6 +57,12 @@ public struct WSError: Error {
     public let code: Int
 }
 
+public enum WebSocketStatus {
```
fassko:
```suggestion
public enum WebSocketStatus: String {
```

How about using String instead of default Int?

lickel: By default, Swift enums aren’t backed by anything. The default isn’t raw representable at all. 

With that in mind, do you still want this change?

lickel:
@fassko I left a comment above. Mind taking a look?

acmacalister:
The codebase has actually under gone a major refactor and completely changed how we handle the whole isConnected state. Feel free to check it out here #653. Feedback welcomed.

daltoniam:
I'm going to go ahead and close this out as it is related to the old 3.X.X codebase. The connection state should be handled properly in the 4.0.0 codebase. Thanks for the PR!

PR #590, opened 2018-10-29
Don't add empty Sec-WebSocket-Protocol header

sirlantis:
> The request MAY include a header field with the name
> |Sec-WebSocket-Protocol|.  If present, this value indicates one
> or more comma-separated subprotocol the client wishes to speak,
> ordered by preference.
> 
> -- The WebSocket Protocol (https://tools.ietf.org/html/rfc6455)

Passing an empty / neutral array should be the same as passing `nil` here.
The WebSocket API (in Browsers) also seems to treat `null` and `[]` the same.

Preferably the initializer signature should be `protocols: [String] = []`,
but that would make this a breaking change.


fassko:
Thanks @sirlantis for your contribution!

PR #597, opened 2018-11-09
WSError should print to console if DEBUG

austintt:
Currently, Starscream doesn’t display errors in the console. These changes will print errors when the preprocessor condition `DEBUG` is set.


acmacalister:
You can print errors to the console yourself in the callback or delegate. I don't feel like this is something the library should do for you. Thanks though!

PR #603, opened 2018-12-15
Static queue can cause unexpected timeout when there are multiple ws connections.

cnliusen:
The queue is used to check whether the initiation of a  connection is timeout. When there are two instances of FoundationStream and they are both trying to initiate a connection, if one of them waits too long, the queue will be blocked. As a result the other instance may also timeout. This causes "Timed out waiting for the socket to be ready for a write".


xymtek:
Can we get this fix merged, this is also causing me issues and this fix does work.

fassko:
@cnliusen Thanks for your contribution and nice catch!

PR #607, opened 2019-01-10
update NetwokStream to support wss SSL valid cert

bukira:
change "false" to "true" for isValidSSLCertificate() in NetworkStream
This allows NetworkStream to work with WSS web sockets and valid SSL certs.
Apple automatically checks certs so no manual check is required

### DIFF
--- a/NetworkStream.swift
+++ b/NetworkStream.swift
@@ -1,0 +1,142 @@
+//////////////////////////////////////////////////////////////////////////////////////////////////
+//
+//  NetworkStream.swift
+//  Starscream
+//
+//  Created by Dalton Cherry on 7/23/18.
+//  Copyright (c) 2014-2018 Dalton Cherry.
+//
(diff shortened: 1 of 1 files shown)


fassko:
@bukira you need to edit existing file in Sources/Starscream/NetworkStream.swift

Right now you have added that file in home directory.

bukira:
@fassko 

Yes but I have no idea how to do that, I gave it a go and obviously it wasn't right, as I have no idea how to do what your asking  I can't do.

All I request is 1 word change from "false" to "true"

fassko:
@bukira you need to clone xcode-10 branch and edit file Sources/Starscream/NetworkStream.swift

bukira:
Thats exactly what I did

fassko:
If you check files changed then you can see that NetworkStream.swift has moved to home dir

acmacalister:
See PR #653. It refactored all this logic and has certificate pinning support.

PR #609, opened 2019-01-15
verify OSStatus returned from Sec* functions

michaelkirk-signal:


daltoniam:

PR #614, opened 2019-01-25
Fix code signing

c0diq:
There’s no need to define code signing for frameworks. When hardcoding iOS Developer, this would prevent this framework from getting signed for Enterprise for example.

Also fixed Unit Tests which didn't build on Xcode 10.1


c0diq:
Anybody?

fassko:
Thanks @c0diq for your contribution.

PR #617, opened 2019-02-07
Avoid data race with CompressionState

michaelkirk-signal:
This issue was found by enabling TSan, then receiving and sending lots of messages concurrently.

Handling reads is serialized on one queue, and handling writes is serialized on
another queue, which means that a read can be handled concurrently with
handling a write.

`websocket.compressionState` is read while handling writes and handling reads.
`websocket.compressionState` is mutated while handling reads.

So it's possible to get a data race where handling a read mutates `websocket.compressionState` while handling a write is trying to read from `websocket.compressionState`.


acmacalister:
Thanks for you PR! The codebase has actually under gone a major refactor and changed the way we handle compression on data. Feel free to check it out here #653. Feedback welcomed.

daltoniam:

PR #618, opened 2019-02-09
Better way to fix archs

c0diq:
This is a better way to fix archs by not specifiying any valid architecture at the target level.
Also fixed some warnings.


fassko:
Thanks @c0diq for your contribution.

PR #626, opened 2019-02-27
objective-c implementation

tolgatanriverdi:
By applying these fixes, we are able to use starscream framework inside our objective-c project as same as we used it in swift project


fassko:
Thanks @tolgatanriverdi I will check it in coming days!

nuclearace:
@tolgatanriverdi The original intention was for this to be a Swift only library. There is https://github.com/acmacalister/jetfire for Obj-C

acmacalister:
@tolgatanriverdi thanks for your PR, but @nuclearace is correct. This is a Swift only library.

PR #631, opened 2019-03-03
Update web sockets org echo example

yamoridon:
* Add a project file for WebSocketsOrgEcho example because it lacks the project file.
* Add a button to send messages to the echo sever. Because I think it strange that an echo client keeps listening the echo server without sending any message.


fassko:
Thanks @yamoridon !

I will review it in coming days.

PR #642, opened 2019-03-28
Remove unnecessary code in WebSocket.swift

texuf:
The previous code was pushing a message into a queue, reading from the queue, and removing the message from the queue all in the same function call. Perhaps this is dead code from a previous implementation? Removing for simplicity and clarity.

If I'm missing something please let me know. 

Test Plan:
I ran all the tests, and used the new code in my application.


acmacalister:
Thanks for you PR! The codebase has actually under gone a major refactor on we process messages. Feel free to check it out here #653. Feedback welcomed!

texuf:
Sounds good! I’ll close this and check out the new branch.

PR #645, opened 2019-04-10
Add compatibility for prior swift versions

PierrePerrin:
Replace all  `guard let self` to `guard let strongSelf`

Add SSCommonCrypto module for swift version prior than 4.2.


acmacalister:
Thanks for you PR! The codebase has actually under gone a major refactor and replaced `guard let self = self`. Feel free to check it out here #653. Feedback welcomed!

daltoniam:

PR #653, opened 2019-04-27
v4 Refactor Notice (WIP)

acmacalister:
This branch is currently a major refactor of the codebase into more modern Swift. It is protocol driven to make things more easily extensible. The code should be stable and ready for testing (it still passes the whole Autobahn test suite). It also makes use of the new `Network.framework` Below are some of the TODOs left:

- [ ] Test WatchOS/TVOS support. The original Core Foundation APIs have been updated to use the Network.framework or Foundation APIs as a fallback. Testing is still needed to see if this will effect WatchOS. In the past, WatchOS didn't have the Foundation APIs available, meaning we were forced to drop down to the CoreFoundation APIs. If this is still the case, Another `Transport` will need to be implemented to bring this functionality back.

- [ ] Finish the tests. Work has been done to replicate all the existing Autobahn test suite. Examples can be seen in the Tests directory. We would like to have all 500+ base cases tests so CI builds could easily test PRs for compatibility. Additional tests for the Certificate/Key Pinning are also planned. This is a huge improvement over the current testing cycle which is generally done manually.

- [ ] Experimental server support. With the addition of the Network.framework, WebSocket server support is much easier and being explored. (See the Server folder for more information)

- [ ] Test Linux support.

Any testing or feedback are appreciated. If any of the above TODOs sound interesting to you, feel free to comment on this PR and we can help get you setup to contribute on those.


fassko:
@acmacalister @daltoniam I have Good news.
I got to know trough some trusted sources that Apple is going to release Network.framework for watchOS.

Regarding this PR I will focus on it later this week or next week.

acmacalister:
awesome to hear @fassko. The `Network.framework` has been really nice to work with.

Sources/Security/FoundationSecurity.swift
```diff
@@ -0,0 +1,101 @@
...
+        
+        if let validateDomain = domain {
+            SecTrustSetPolicies(trust, SecPolicyCreateSSL(true, validateDomain as NSString?))
+        }
+        
+        handleSecurityTrust(trust: trust, completion: completion)
+    }
+    
+    private func handleSecurityTrust(trust: SecTrust, completion: ((PinningState) -> ())) {
+        if #available(iOSApplicationExtension 12.0, *) {
```
dcvz: shouldn't this be `if #available(iOS 12.0, *)`?

fassko:
Apple released today this one https://developer.apple.com/documentation/network/nwprotocolwebsocket

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Sources/Security/FoundationSecurity.swift
acmacalister: It should. Thanks for the heads up

acmacalister:
I watched the Advances in Networking part 1 today, where the Apple engineers discussed the WebSocket additions in URLSession as well the Network.framework. Excited to see the advancements and native support for WebSockets, but that does brings up questions for the future of Starscream. In my mind this brings up a couple of options what we do going forward:

1. Add native support to Starscream in this v4 refactor branch and retain the existing API.
2. Release the v4 refactor branch and create a v5 branch that basically makes Starscream backward compatible for older version of Swift/iOS/MacOS that don't support the new API.
3. Just release the v4 refactor and start the depreciation process of Starscream in favor of the native API.

I think each of these have pros and cons,, but figured I would get the conversation started. Below are my thoughts on each:

1. This is great for people already using Starscream. No breakage, business as usual and very little development work. That being said v4 already has some breaking changes, so now would be the time to strike if we are planning big API changes to match how the native framework works.
2. This would be super cool for pretty much everyone, but could be a huge amount of development work. 
3. Low effort, encourage use of native APIs, but requires developers to have ifdefs in their code to either use Starscream or native APIs if they want to support older version of iOS.

As always, I would love feedback on others thoughts. @daltoniam @fassko if I could get y'all feedback in particular on what option you like that would be great.

daltoniam:
I agree this is an exciting development. After 5 years of development on Starscream, we are getting a native WebSocket API. Excitement and sentiment aside, I think we should add the native API support if it is available. It is going to be the most work, but I don't see a reason to not use the native APIs "under the hood" when available, especially since we are already refactoring everything. I think this will allow for the easiest transition from Starscream to the native APIs as it makes sense and for easy support of the different API availability on each OS version.

fassko:
My 2 cents would be to release current branch now and start working to support native APIs behind the scene then. Basically plan nr. 2 from @acmacalister comment https://github.com/daltoniam/Starscream/pull/653#issuecomment-499738690 

I'm actually very happy that Apple finally is bringing native support for WebSockets in Network.framework and URLSession. We need to support -1 or sometimes -2 iOS version for our apps so it will take to adapt it fully and in either way Starscream will be used at least 2 years from now. But assume it will be even longer.

rudyrichter:
I agree with that, we have some changes we'd like to submit that we're holding off on until this PR is merged.

rudyrichter:
@fassko any movement here?

fassko:
@rudyrichter sadly no, but soon :)

WyattMufson:
@fassko any guess when this will be finished? Thanks

daltoniam:
@acmacalister and @fassko I think I finished up everything so we can (finally) get this merged!

fassko:
Thanks @daltoniam I will check this in coming days! Sorry for delays. :(

fassko:
@daltoniam can you take a look at unit tests? I can't build them. 
image

Marcocanc:
This is amazing.
Just saw that the `URLSessionWebSocketTask` stuff is commented out though.
Any idea when this will be merged?

sjmueller:
I agree, excited that starscream will use the native `URLSessionWebSocketTask` api. Can't wait to see this merged.

PR #655, opened 2019-05-02
Fix crash when trying to connect with invalid url

ggarber:
When using an invalid url (f.e. ",") and trying to connect it crashes when trying to access the scheme with `url.scheme!`.  With this change if the scheme or host is missing it will stop trying to connect in the same way it is done when request.url is not available.


daltoniam:

PR #665, opened 2019-06-04
Fix parsing HTTP responce header values which contains colon symbol

OlehKulykov:
If HTTP value for header contains ":" character, than prev. implementation will take only first part of the value till ":" char and ignore everything else.
Example: "Date: Tue, 04 Jun 2019 13:55:20 GMT" => headers["date"] = "Tue, 04 Jun 2019 13"

Please accept this bug fix and provide new release.
Thanks for considering my request.


daltoniam:
I added this in 4.0.2. Thanks for the PR!

PR #651, opened 2019-04-26
Prevent writing to socket if not connected

2h4u:
We have this small change in production since 2 months and since then we didn't encountered the crash discussed in #588 anymore.

PR #670, opened 2019-06-20
Always wait for currently executing operations to finish

mgrant34:
We should always be waiting until currently executing operations have completed before we set outputStream to nil, and avoid another concurrentOperation from accessing a nil outputStream.

Per the documentation on `cancelAllOperations`: https://developer.apple.com/documentation/foundation/nsoperationqueue/1417849-cancelalloperations

```Canceling the operations does not automatically remove them from the queue or stop those that are currently executing```

### DIFF
--- a/Sources/Starscream/WebSocket.swift
+++ b/Sources/Starscream/WebSocket.swift
@@ -723,11 +723,11 @@ open class WebSocket : NSObject, StreamDelegate, WebSocketClient, WSStreamDelega
      Disconnect the stream object and notifies the delegate.
      */
     private func disconnectStream(_ error: Error?, runDelegate: Bool = true) {
-        if error == nil {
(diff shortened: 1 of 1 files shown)


mgrant34:
The onus for this change is a persistent but random crash on L241 of WebSocket.swift caused by output stream being nil https://github.com/daltoniam/Starscream/blob/a2ed45c0b2f996cb8c335c4f270ecc68c3bd4c0f/Sources/Starscream/WebSocket.swift#L241

colbyAiken:
good call 👍

rudyrichter:
This looks right to me, but it also looks like it duplicates logic in https://github.com/daltoniam/Starscream/pull/651 ?

mgrant34:
This prevents the issue at the source rather than having if !connected checks everywhere

daltoniam:

PR #683, opened 2019-08-05
Add cookies support

kilyukhin:
Since WebSocket handshake is an HTTP request, server and client may exchange HTTP cookies according to RFC 6455.


daltoniam:
I added this in 4.0.2. Thanks for the PR!

PR #691, opened 2019-09-06
replace deprecated methods.

cntrump:


daltoniam:
I'm going to go ahead and close this out as it is related to the old 3.X.X codebase. The deprecated methods calls should be resolved. Thanks for the PR!

PR #697, opened 2019-09-24
Fix for a buffer overflow when close code has not been received from the socket

mikaryyn:
I encountered the crash here where `pointer` only contained two bytes, [136, 39] and it was trying to read `closeCode` from outside the buffer. This should fix it.


daltoniam:
Thanks!

PR #711, opened 2019-10-25
Using `reduce(into:,_)` instead of `reduce(_,_)`

valeriyvan:
In this use case it's better because it doesn't produce intermediary arrays on every step. And it's more concise.


daltoniam:
I'm going to go ahead and close this out as it is related to the old 3.X.X codebase. 4.0.0 removed the rudimentary SSL pinner in favor of a protocol that allows custom support. Thanks again for the PR!

PR #712, opened 2019-10-25
Get rid of all mentioning of NSData

valeriyvan:


daltoniam:
I'm going to go ahead and close this out as it is related to the old 3.X.X codebase. The 4.0.0 codebase should have removed these old calls. Thanks again for the PR!

Issue #748, opened 2020-03-11
Cannot connect to secure wss:// websocket on iOS 11 & earlier


nickdowell:
The `FoundationTransport` (used on iOS 11 & earlier) does not establish a secure connection to the host, resulting in a failed connection to hosts that require security.

The following code snippet replicates the problem on all platforms by forcing use of the `FoundationTransport`

```swift
import Starscream

class WebsocketTester: WebSocketDelegate {
    
    var webSocket: WebSocket?
    
    func start() {
        let url = URL(string: "wss://echo.websocket.org")!
        webSocket = WebSocket(request: URLRequest(url: url), engine:
            WSEngine(transport: FoundationTransport(), certPinner: FoundationSecurity()))
        webSocket?.delegate = self
        webSocket?.connect()
    }
    
    func didReceive(event: WebSocketEvent, client: WebSocket) {
        print(#function, event)
    }
}
```

The output of running this is

```
didReceive(event:client:) cancelled
```

This has previously been reported in https://github.com/daltoniam/Starscream/issues/728

jaltreuter:
@nickdowell thank you for putting up that pull request.  This same issue started happening in my app after upgrading from Starscream 3.0.5 to 4.0.0.  I ended up simply making a copy of FoundationTransport and adding your changes, and that was effective at solving the problem.

I'd love to see this get some attention from the devs, since it's easily reproducible and affects a reasonably large audience.

Issue #728, opened 2020-01-22
WebSocket instantly disconnect on iOS 10 using Starscream 4.0.0


p-rob:
Starscream: 4.0.0
iOS: 10.3.4
Devices: iPhone 5 (MD299F/A) - iPad pro (9.7 inches; simulator) 

When connecting to a secure web-socket (wss:// or https://) the framework instantly send a `.cancelled` event coming from `FoundationTransport.stream(_ aStream: Stream, handle eventCode: Stream.Event)` `.endEncountered` event.

After further investigation, it looks like `FoundationTransport.getSecurityData()` always return nil.

martinpristas:
Same here

PR #749, opened 2020-03-11
Enable SSL in FoundationTransport for secure hosts

nickdowell:
This fixes connection to secure hosts such as wss://echo.websocket.org

The `kCFStreamPropertySocketSecurityLevel` property needs to be set for the CFStream(s) to negotiate a secure connection with the server.

- Fixes https://github.com/daltoniam/Starscream/issues/748
- Fixes https://github.com/daltoniam/Starscream/issues/728

### DIFF
--- a/Sources/Transport/FoundationTransport.swift
+++ b/Sources/Transport/FoundationTransport.swift
@@ -66,6 +66,12 @@ public class FoundationTransport: NSObject, Transport, StreamDelegate {
         inStream.delegate = self
         outStream.delegate = self
     
+        if isTLS {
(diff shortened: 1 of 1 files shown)


michalcichon:
Hey @nickdowell, can you paste here what kind of error do you get because of this bug?

nickdowell:
Hey @michalcichon the bug causes connections to secure hosts to fail with the following:

```
WebSocketEvent.cancelled
```

Sample code for reproducing the issue was included in #748

daltoniam:
Thanks!

PR #759, opened 2020-04-08
Enum updated.

itsaboutcode:


daltoniam:
I think `viability` is correct. `viablity` was a misspelling in the code. Thanks for the PR though.

PR #764, opened 2020-04-16
Fix the deadlock issue on iOS 11

ameyjain:
If the swift version is less than 5.0 instead of calling the default function: https://developer.apple.com/documentation/swift/2632391-withunsafemutablebytes 
it calls itself causing a deadlock.

### DIFF
--- a/Sources/DataBytes/Data+Extensions.swift
+++ b/Sources/DataBytes/Data+Extensions.swift
@@ -27,22 +27,20 @@ import Foundation
 internal extension Data {
     struct ByteError: Swift.Error {}
     
+    #if swift(>=5.0)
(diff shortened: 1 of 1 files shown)


jagdeep-manik:
+1, I ran into the same issue. It'll cause an infinite loop. @daltoniam 

As an aside, the way swift package manager is set up makes it so that it builds using swift 4.2 even though we could technically support swift 5. I wish there was a way to tell SPM to build with this with swift 5. 😬

I'm also wondering if we need those version checks in the first place--it looks like all those functions in those helpers are available in swift 4. Does swift 5 require some different memory management that forces us to have special handling?

amayers:
The reason that SPM is building it with 4.2, is because this line at the top of `Package.swift`.

`// swift-tools-version:4.2`

daltoniam:
Thanks! Appreciate the PR!

PR #766, opened 2020-04-21
Fix deprecation warning in WSCompression

nickoto:
Apple recommends using the withUnsafeMutableBytes method on the buffer
to bind an unsafe pointer.


rudyrichter:
@daltoniam this looks like a good change to merge.

daltoniam:
Thanks!

PR #772, opened 2020-05-06
feat(waiting): add error handling for errors while waiting

DomiR:
While using a custom cert pinnner calling `completion(.failed(error))` will update the state with state `waiting` passing the cert error. 

In this case we also should call the error handler.


DomiR:
What do you think about this?

daltoniam:
Makes sense, I like it. Going to cut another version (to fix some bugs) before merging this since it is a breaking changing with the enum case change. Thanks for the PR!

scoreyou:
@daltoniam Are there any updates on this?

PR #775, opened 2020-05-12
Remove tls handling override when no pinner provided

myeyesareblind:
When no pinner was provided, it is expected that framework will use
whatever iOS will provide. But instead - it will allow ALL
connections, bypassing any verification.
This is breaking change, but most clients of the library probably
never had to do that.
Client can still allow selfSigned certs with FoundationSecurityError.allowSelfSigned.

### DIFF
--- a/Sources/Transport/TCPTransport.swift
+++ b/Sources/Transport/TCPTransport.swift
@@ -59,13 +59,9 @@ public class TCPTransport: Transport {
         options.connectionTimeout = Int(timeout.rounded(.up))
 
         let tlsOptions = isTLS ? NWProtocolTLS.Options() : nil
-        if let tlsOpts = tlsOptions {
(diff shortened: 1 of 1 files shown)


myeyesareblind:
The problem I encountered in the first place is:
Wildcard certificates are not validated properly.

I couldn't understand how to fix that particular issue, so looked for a workarounds.
The workaround with pinner == nil looks dangerous to me.

ghost:
Hi  @daltoniam @acmacalister,

As part of ongoing research, we found that the current behavior of Starscream does not use the normal iOS certificate verification logic if pinners are not provided. As a result, the default configuration is vulnerable to MITM attacks using, for example, valid TLS certificates for the wrong domain. This PR fixes the issue by not overriding the verifier if a pinner is not provided.

Could you please take a look at this PR? We believe it constitutes a vulnerability.

Issue #773, opened 2020-05-06
[CRASH] CoreFoundation _inputStreamCallbackFunc


nickoto:
I did a little bit of investigation into this issue ( #676 ) since I encountered it. If the application releases the WebSocket instance before the disconnect is completed the `FoundationTransport` is released but it has not yet been stopped. So messages (from 'CFReadStream'/'CFWriteStream') to their delegates will still get fired and will result in this crash.

I'm not sure what the right solution is to the problem; however, `deinit` in `FoundationTransport` to do the right cleanup or ensuring that engine doesn't release the transport until it is finished stopping (since the stop is currently async, and only weakly holds onto self in the completion handler).

Otherwise the client needs to wait until the socket has been completely closed before releasing its instance (which would need docs -- but I'd much prefer it doesn't crash if you release the webSocket object).

A simple test for this problem:

Connect a websocket and then...
```
        webSocket.delegate = nil
        webSocket.disconnect()
        webSocket = nil
```

shalyf:
I fixed it with these (#777) , thanks
```swift
deinit {
    if #available(iOS 12, *) { } else {
        if let mirror = Mirror(reflecting: self).superclassMirror {
            if let inputStream = mirror.descendant("engine", "transport", "inputStream") as? Stream {
                inputStream.delegate = nil
            }
            if let outputStream = mirror.descendant("engine", "transport", "outputStream") as? Stream {
                outputStream.delegate = nil
            }
        }
    }
}
```

jaltreuter:
@shalyf that implementation seems very fragile.  What's the reason behind all those extra checks?

Given that `inputStream` and `outputStream` are both private, and the implementation of `disconnect()` sets their delegates to `nil` without any check like you've done, I think it's safe to take the simpler approach I implemented in my #777 solution.  I'm open to hearing other opinions though.

shalyf:
@jaltreuter I have MyWebSocket inherited from WebSocket and implement deinit to fix this crash. My code is based on your solution, use Mirror to access private property, because I don't want to modify the framework. Before the author updates, this is my temporary code.

aserdobintsev:
@shalyf thanks for the idea about a temporary solution, based on @jaltreuter PR 👍

philfi:
[...]

Issue #676, opened 2019-07-17
[CRASH] CoreFoundation _inputStreamCallbackFunc


bobbyrehm:
This is on version 3.1.0.

Here's the stack trace via Crashlytics:

Crashed: com.vluxe.starscream.websocket
EXC_BAD_ACCESS KERN_INVALID_ADDRESS 0x0000000edb473fd0

0  libobjc.A.dylib                0x1aadfb530 objc_msgSend + 16
1  libobjc.A.dylib                0x1aaddf22c -[NSObject respondsToSelector:] + 36
2  CoreFoundation                 0x1abb8276c _inputStreamCallbackFunc + 56
3  CoreFoundation                 0x1abbadbc8 _signalEventSync + 216
4  CoreFoundation                 0x1abbadc28 ___signalEventQueue_block_invoke + 24
5  libdispatch.dylib              0x1ab648a38 _dispatch_call_block_and_release + 24
6  libdispatch.dylib              0x1ab6497d4 _dispatch_client_callout + 16
7  libdispatch.dylib              0x1ab5f2320 _dispatch_lane_serial_drain$VARIANT$mp + 592
8  libdispatch.dylib              0x1ab5f2e3c _dispatch_lane_invoke$VARIANT$mp + 428
9  libdispatch.dylib              0x1ab5fb4a8 _dispatch_workloop_worker_thread + 596
10 libsystem_pthread.dylib        0x1ab829114 _pthread_wqthread + 304
11 libsystem_pthread.dylib        0x1ab82bcd4 start_wqthread + 4

arnold-plakolli:
Happening to me as well. Same stack trace

rztrom:
I have this problem as well. Same stack trace

bobbyrehm:
I'm still seeing this crash after reverting to 3.0.6.

yan-zaitsev:
I see same crashes on 3.1.0

OlehKulykov:
I'm also conforming the same crash & stack trace, crashing up to ~10-20 times per day.

fassko:
Dealing with these issues is pain in the ass. I hope soon we all can use URLSession Websocket implementation. :)

daltoniam:
[...]

PR #777, opened 2020-05-13
In FoundationTransport deinit, remove self as a delegate to InputStream and OutputStream to prevent EXC_BAD_ACCESS

jaltreuter:
As mentioned by @nickoto in #676 and #773, Starscream will crash if a socket is deallocated before it finishes disconnecting when using `FoundationTransport`.  In my own testing, simply setting the delegates to `inputStream` and `outputStream` to `nil` in `deinit` is effective at preventing the crash.  This makes a fair amount of sense, since [`Stream.delegate` is an `unowned` property](https://developer.apple.com/documentation/foundation/stream/1418423-delegate).

### DIFF
--- a/Sources/Transport/FoundationTransport.swift
+++ b/Sources/Transport/FoundationTransport.swift
@@ -47,7 +47,12 @@ public class FoundationTransport: NSObject, Transport, StreamDelegate {
         onConnect = streamConfiguration
     }
     
-     public func connect(url: URL, timeout: Double = 10, certificatePinning: CertificatePinning? = nil) {
+    deinit {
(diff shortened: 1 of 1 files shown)


daltoniam:
Good catch, Thanks! Appreciate the PR.

alexeybolv:
@daltoniam Hi, when are you plan to make release with this fix? I havу too many crashes in my product app((
Thank you!

Vactorwang:
> @daltoniam Hi, when are you plan to make release with this fix? I have too many crashes in my product app((
> Thank you!

Same

Issue #786, opened 2020-06-04
Infinite loop in TCPTransport.swift : readLoop()


sdidla:
**Scenario:**
1. Simple WebSocket server running locally
2. Launch `SimpleTest.xcodeproj` to connect to `ws://localhost:8080`
3. Close the `NWConnection` on the server

**Result:**

On this [line](https://github.com/daltoniam/Starscream/blob/master/Sources/Transport/TCPTransport.swift#L141) : 

```swift
data = nil
error = nil
```

 thread `com.vluxe.starscream.networkstream (serial)` enters infinite loop eating up CPU. looking for one more frame even though `context.isFinal = true`

horizon554:
facing same problem as you

sdidla:
Replacing [line 144](https://github.com/daltoniam/Starscream/blob/master/Sources/Transport/TCPTransport.swift#L144) with:
```swift 
if data != nil || context?.isFinal == false {
    s.readLoop()
}
```
works for me. Is this a possible solution?

horizon554:
Figured it out:
Close socket when your heartbeat not get response

sdidla:
But you will still have the infinite loop (and high CPU usage) till the heartbeat timeout, right?

fruitcoder:
I think this can be marked as fixed right @sdidla?

sdidla:
✅

PR #790, opened 2020-06-06
Improve conditions for calling readLoop() and prevent infinite looping

sdidla:
This is a fix for https://github.com/daltoniam/Starscream/issues/786 . 

Taken from Apple's sample code for `Network.framework` https://developer.apple.com/documentation/network/implementing_netcat_with_network_framework 

Specifically:
```c
dispatch_block_t schedule_next_receive = ^{
    // If the context is marked as complete, and is the final context,
    // we're read-closed.
    if (is_complete &&
        context != NULL && nw_content_context_get_is_final(context)) {
        exit(0);
    }

    // If there was no error in receiving, request more data
    if (receive_error == NULL) {
        receive_loop(connection);
    }
};
```

### DIFF
--- a/Sources/Transport/TCPTransport.swift
+++ b/Sources/Transport/TCPTransport.swift
@@ -141,7 +141,16 @@ public class TCPTransport: Transport {
             if let data = data {
                 s.delegate?.connectionChanged(state: .receive(data))
             }
-            s.readLoop()
+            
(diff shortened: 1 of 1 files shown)


daltoniam:
Thanks! I have a vague memory of doing some code that checked something like this, but I can't remember why it didn't make it in this class. Thanks again for the PR.

WTN-M:
www.google.com

PR #797, opened 2020-06-17
Switch SPM to use Swift 5.2

amayers:
Since if you install Starscream via Cocoapods, it will use Swift 5.0, and the project file also uses 5.x. So there doesn't appear to be a reason for SPM to be limited to 4.2.


daltoniam:
Thanks! Appreciate the PR!

PR #799, opened 2020-06-22
Fix compile error on Xcode 12

amayers:
Now you can build on both Xcode 11.5 & 12.0 b1.

On 12, the unit tests would fail to compile because it wasn't finding the `Starscream` framework. So I added `Starscream` as a dependency of the test target.

In `FoundationHTTPServerHandler` the `method` is of type `CFString`, while `getVerb` is `NSString`. Looks like Xcode 12 doesn't like equality comparison between those types without a cast as `NSString`.

### DIFF
--- a/Sources/Framer/FoundationHTTPServerHandler.swift
+++ b/Sources/Framer/FoundationHTTPServerHandler.swift
@@ -74,7 +74,7 @@ public class FoundationHTTPServerHandler: HTTPServerHandler {
             return false //not enough data, wait for more
         }
         if let method = CFHTTPMessageCopyRequestMethod(response)?.takeRetainedValue() {
-            if method != getVerb {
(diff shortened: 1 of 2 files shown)


austtn:
Can someone merge this?

amayers:
TBH the slow response to this simple PR caused me to reevaluate our needs, and I found that `URLSessionWebSocketTask` exists in iOS 13+ and is sufficient for our needs (originally we supported iOS 12). So I have removed this library from our app.

I totally get that it is likely a side project for someone, and so they must prioritize accordingly.

rudyrichter:
@daltoniam the code change here looks good

daltoniam:
Thanks!

y25zhao:
Hey guys, shouldn't we bump up the podspec version for this fix? Thanks

Issue #719, opened 2019-12-27
Warning 'couldn't find pc file' when using SwiftPM


nkristek:
When importing Starscream version 3.1.1 (currently latest) using SwiftPM I get the warning `couldn't find pc file`. I tried version 3.1.1 and master branch, when using 3.1.0, the warning does not appear.

## Steps to reproduce
- create new iOS project (File > New > Project > iOS > Single View App)
- add SwiftPM dependency (File > Swift Packages > Add Package Dependency...), select either exact version 3.1.1 or branch master
- build

patburns:
Also have this warning. It's because of this issue in swift-nio-zlib-support:

https://github.com/apple/swift-nio-zlib-support/issues/3

Of which starscream currently depends on.

patburns:
I commented in the other two threads. I don't have the context on why streamscreeam uses swift-nio-zli-support, but you can "silence" this warning by installing pkg-config on the device that you are compiling on:

`brew install pkg-config`

This really isn't a fix but a workaround. Maybe it should be considered adding this to the documentation for this project? Or maybe evaluated how swift-nio-zli-support is being used?

nkristek:
Thank you for that workaround

yspreen:
Two and a half months later and it's still an issue... Do I really have to add this to the common issues page of our project? Or is there a proper fix?

yspreen:
Since 3.1.0 does not have this issue, is there any crucial functionality missing from the latest version>

nkristek:
This issue seems to come from swift-nio. Thus I will close this issue, thanks!

PR #807, opened 2020-07-12
Conditionally include 'swift-nio-zlib-support' dependency to avoid warning on macOS

varahash:
This PR removes an ugly Xcode warning (https://github.com/daltoniam/Starscream/issues/719)
image


daltoniam:
👍

headlessme:
@daltoniam any chance we can get this backported to 3.x? We need a warning-free build and can't upgrade to version 4 yet due to some outstanding issues

PR #812, opened 2020-07-28
Allow WebSocketDelegate 'didReceive' method's 'client' parameter to work with `WebSocketClient` mock objects

### LINKED ISSUE #811: Allow WebSocketDelegate 'didReceive' method's 'client' parameter to work with 'WebSocketClient' mock objects
Unless there's a technical reason that this can't or shouldn't be the case, I think that the method signature of the `WebSocketDelegate` method should be changed from:

```swift
func didReceive(event: WebSocketEvent, client: WebSocket)
```
to:
```swift
func didReceive(event: WebSocketEvent, client: WebSocketClient)
```
Since that would allow a mock of a `WebSocket` used for unit testing purposes to return itself if that mock calls this delegate method. This seems sensible since I can see that the `WebSocketClient` protocol seems to have primarily been introduced to solve the problem of creating mocks for unit tests.

Is this a change that could be made for a 4.0.4 release? I'm asking as I've recently been upgrading from 3.1.1 to 4.0.3, and previously we were subclassing `WebSocket` and overriding some relevant methods, but this is no-longer possible since they've changed from `open` to `public`. Therefore I am migrating this our mock to conform to `WebSocketClient` instead. However, I've run into this issue because we call `didReceive(…)` in a few places and it's expecting a concrete `WebSocket` currently.

JonathanDowning:
Resolves #811

### DIFF
--- a/Sources/Starscream/WebSocket.swift
+++ b/Sources/Starscream/WebSocket.swift
@@ -88,7 +88,7 @@ public enum WebSocketEvent {
 }
 
 public protocol WebSocketDelegate: class {
-    func didReceive(event: WebSocketEvent, client: WebSocket)
+    func didReceive(event: WebSocketEvent, client: WebSocketClient)
 }
 
 open class WebSocket: WebSocketClient, EngineDelegate {


danielrbrowne:
@daltoniam Do you require anything else from me in order for this change to be merged in and released?

JonathanDowning:
> @daltoniam Do you require anything else from me in order for this change to be merged in and released?

No, I'm happy to merge this :)

daltoniam:
Thanks and sorry about the delay. It will be somewhat of a breaking change, but I agree it should have been like that from the start. Thanks again!

machl:
@acmacalister This was released with [4.0.6](https://github.com/daltoniam/Starscream/releases/tag/4.0.6) but contains breaking changes to public API. It causes issues to many libraries which uses this as a dependency.
Next time please consider bumping minor or major version instead.

acmacalister:
@machl, yes apologizes. It should have been a minor version bump instead of a patch.

nunogoncalves:
Hi! Sorry, following semver, it should have been a major instead of a minor because it breaks clients codes.

PR #821, opened 2020-08-17
fix waiting error for TCPTransport

Ranqb:
when using `NWConnection`, if there are problems accessing the socket and the server breaks the connection when requesting a handshake, `stateUpdateHandler` return case `.waiting`
There is no handler for this case and we don't get an event in `WebSocketDelegate`

In my case returns the following errors:

```
 [BoringSSL] nw_protocol_boringssl_input_finished(1543) [0x7ff02ec0a0a0] Peer disconnected during the middle of a handshake. Sending errSSLClosedNoNotify(-9816) alert
 [] nw_http_connect_process_response http connect Proxy received status: HTTP/1.0 503 Error
 [BoringSSL] nw_protocol_boringssl_error(1584) [0x7ff02ec0a0a0] Lower protocol stack error: 57
 [] nw_protocol_http_connect_disconnected http connect disconnected before negotiation complete
 [] nw_endpoint_proxy_start_next_child [C13 tst.ru:443 in_progress proxy (satisfied)] no child at index 2
```


Ranqb:
@daltoniam what decision will be made?

aChase55:
Any good reason as to why this hasn't been merged?

kishanios123 (approved):

fukemy:
hello, i still got problem when using ios 16

PR #845, opened 2020-10-21
Remove three unused properties

XueshiQiao:
These three properties in `WebSocket` class are never used, so maybe it's better to remove them

```swift
    private let writeQueue = DispatchQueue(label: "com.vluxe.starscream.writequeue")
    private var canSend = false
    private let mutex = DispatchSemaphore(value: 1)
```


daltoniam:
Nice! Good catch, must have forgot to remove them when I moved it into the engine class. Thanks again!

PR #854, opened 2020-12-10
Update Starscream.podspec

uypanha:
Fixed Xcode 12 iOS minimum version support


acmacalister:
thanks. We went ahead and updated it today.

pabiagioli:
you missed updating the spec.version field
I'm having issues since 4.0.5 like this
```
dyld[799]: Symbol not found: _$s10Starscream17WebSocketDelegateP10didReceive5event6clientyAA0bC5EventO_AA0bC0CtFTq
```
Did you modify the WebSocketDelegate signatures? maybe the podspec version field is related to this error

I'm using the pod like this
```
pod 'Starscream', '~> 4.0.0'
```
it has been solved when changing the line like this
```
pod 'Starscream', :git => 'git@github.com:daltoniam/Starscream.git', :tag => '4.0.4', :submodules => true
```
versions `4.0.5` and `4.0.6` say I'm installing `4.0.4`

PR #863, opened 2021-01-25
Add Headers to `notAnUpgrade` error.

hishnash:
Some Websocket servers require Auth on the HTTP connection. They will use the WWW-Authenticate header to indicate why this fails. 

The WWW-Authenticate header is used to includes information about what token/scope you need to acquire.  Due to this it is important that if the WS connection fails to connect that the headers are included in that error so that the application can determine the nature of the authentication errors (eg token expired, or wrong scope etc). 

For example:
```http
     HTTP/1.1 401 Unauthorized
     WWW-Authenticate: Bearer realm="example",
                       error="insufficient_scope",
                       scope="USEast",
                       error_description="Requires scope USEast to access."
```

Other use cases of the response header include error Tracing IDs these are very useful if the server responses with an error (500) if we are able to capture the tracing ID when the user files a report in the application it can be linked to the server side stack trace.

### DIFF
--- a/Sources/Framer/FoundationHTTPHandler.swift
+++ b/Sources/Framer/FoundationHTTPHandler.swift
@@ -76,12 +76,6 @@ public class FoundationHTTPHandler: HTTPHandler {
             return false //not enough data, wait for more
         }
         
-        let code = CFHTTPMessageGetResponseStatusCode(response)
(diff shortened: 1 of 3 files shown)


daltoniam:
Ah yeah, forgot we had this in version 3 😬. Thanks for adding it back in.

Issue #779, opened 2020-05-17
enableCompression does not exist


IsaiahJTurner:
It used to, but it doesn't anymore. It's in the docs.

KimiChiu:
Same problem here. Where is the property? There's no enableCompression in WebSocket nor WSCompression.

daltoniam:
The docs should probably be better worded, but the example should how do accomplish compression support: https://github.com/daltoniam/Starscream#compression-extensions

just add:
```
let compression = WSCompression()
let socket = WebSocket(request: request, compressionHandler: compression) // add the compress object when creating a new Websocket
```

This was changed in 4.0.0 as it allows for custom compression implementations. I'll update the docs to make this clearer.

PR #883, opened 2021-04-19
Fixed documentation on compression

jschmid:
Issue #779 marked the document as outdated, but it was never fixed. Here is the updated README.


acmacalister:
Thank you!

Issue #895, opened 2021-06-27
Minimum iOS version needs to increase


ddaddy:
This library fails to build for device (It builds for simulator fine) in Xcode 13 (BETA 1&2).
The error is:

> Failed to build module 'Network'; this SDK is not supported by the compiler (the SDK is built with 'Apple Swift version 5.5 (swiftlang-1300.0.20.4 clang-1300.0.21.1)', while this compiler is 'Apple Swift version 5.5 (swiftlang-1300.0.20.104 clang-1300.0.21.1)'). Please select a toolchain which matches the SDK.

This error goes away when you set the minimum iOS version to 11.0

rizwan95:
@daltoniam  @ddaddy   I have created a PR and submitted it as well. Here is the link -> https://github.com/daltoniam/Starscream/pull/897

Meanwhile if you want to use the library until the PR is accepted, you can do it by following the below given instructions.

Cocoapods - pod 'Starscream', '~> 4.0.0', :git => 'https://github.com/rizwan95/Starscream.git'

SPM - 

dependencies: [
    .Package(url: "https://github.com/rizwan95/Starscream.git", majorVersion: 4)
]

sdc-78:
I don't understand the link between the error message and the minimum iOS version. Can someone explain to me?
Isn't that just an Xcode bug?

ddaddy:
Yes it could be an Xcode bug, and I have filed a bug report with Apple (FB9296596)
If it gets fixed, great. If not, then the minimum iOS version will need to be increased.

rizwan95:
According to Apple, it is not a bug, but it is a feature. Check this link https://github.com/SwiftKickMobile/SwiftMessages/issues/471#issuecomment-879971936 for more information about the issue.

ddaddy:
Although it's a _feature_ to not allow the availability check, I'd say it's a bug that they have an availability check in the Network framework.

ddaddy:
My bug report suggested this might be fixed in Beta 5, but it's not 🙄

baveku:
> My bug report suggested this might be fixed in Beta 5, but it's not 🙄

Yes, I try but it's not, Do you have any idea to fix it?

ddaddy:
Use a minimum iOS version of 11.0, or don't use Xcode 13 beta's.

ddaddy:
This seems to be fixed in Xcode 13 Release Candidate :)

baveku:
[...]

PR #897, opened 2021-07-06
Increased the iOS deployment target to 11.0 

rizwan95:
### Issue Link 🔗
https://github.com/daltoniam/Starscream/issues/895

### Goals ⚽
This PR increases the minimum deployment target of iOS to 11.0 

### Implementation Details 🚧
If the minimum deployment target is less than 11.0 the project will not build with Xcode 13.0 and Swift 5.5 so the minimum deployment target has been increased.


acmacalister:
Thanks. We went head and updated this today with some other fixes.

Issue #913, opened 2021-08-17
Replace SHA1 usages with SHA512


1oo7:
### What do you want to happen?

Replace SHA1 usage with a more secure algorithm.

### What happens now?

https://github.com/daltoniam/Starscream/blob/df8d82047f6654d8e4b655d1b1525c64e1059d21/Sources/Security/FoundationSecurity.swift#L92-L99

SHA1 is insecure. Use something better.

### Demo Code
> Any demo code that may used to implement/use the desired feature. 

### Describe alternatives you've considered

I would recommend CC_SHA512

### Additional context

China, Russia, N. Korea, Iran...

acmacalister:
https://datatracker.ietf.org/doc/html/rfc6455#section-10.8

RFC sets it as a SHA1

PR #914, opened 2021-08-17
Improve security

1oo7:
### Issue Link 🔗

https://github.com/daltoniam/Starscream/issues/913

### Goals ⚽

Improve security by using SHA512 instead of SHA1, and using an obfuscated string to store the salt instead of storing it in plaintext. 

### Implementation Details 🚧

This PR is to address an issue found as a result of an internal security audit at conducted within the company I work for. 

Incidentally I also updated the usage of `data.withUnsafeBytes` to no longer use the deprecated method that was previously being used by Starscream. That older usage was deprecated in Swift 5.0. I chose the solution used in this PR from [this stackOverflow answer](https://stackoverflow.com/a/55499502/12587089), since it seemed like the safest option. 

Open to suggestions if anyone has an idea.


acmacalister:
https://datatracker.ietf.org/doc/html/rfc6455#section-10.8

RFC sets it as a SHA1

PR #923, opened 2021-09-21
Fix a leak with the TCPTransport

paul-verkada:
### Issue Link 🔗
N/A

### Goals ⚽
Fix memory leak

### Implementation Details 🚧
Disconnect on deinit


acmacalister:
Thanks!

PR #930, opened 2021-10-19
Fix connection race when using WSEngine

### LINKED ISSUE #791: Masked and rsv data is not currently supported code 1002
Hi, I was debugging normaly testing some parsing and now I'm getting this error on websocket.connect() I did a rollback on my minor changes and the error persists.

this is the log:
viabilityChanged
error Optional(Starscream.HTTPUpgradeError.notAnUpgrade(503))
error Optional(Starscream.WSError(type: Starscream.ErrorType.protocolError, message: "masked and rsv data is not currently supported", code: 1002))

yoheimuta:
### Issue Link 🔗

Fixes https://github.com/daltoniam/Starscream/issues/791

### Goals ⚽

WSEngine is susceptible to connection race which causes an upgrade failure showing inexplicable errors like`notAnUpgrade` and `masked and rsv data is not currently supported.` 

In our environment running the [gorilla/websocket server](https://github.com/gorilla/websocket), we found out that the corresponding server-side error was ["websocket: client sent data before handshake is complete"](https://github.com/gorilla/websocket/blob/master/server.go#L185). That's because it's a security model violation mentioned in [RFC 6455](https://datatracker.ietf.org/doc/html/rfc6455#section-1.6).

It would be nice if the library strictly conforms to a WebSocket's security model without users' extra hassle. Because full-blown apps could easily run into the race condition, as they implement a thorough reconnection mechanism triggered at different timing.

### How to reproduce 📝 

The most straightforward code is below.

```swift
socket = WebSocket(request: request)
socket.connect()
socket.connect() # <- Add this line.
```

Say the first connect() creates a [TCPTransport's connection](https://github.com/daltoniam/Starscream/blob/master/Sources/Transport/TCPTransport.swift#L33) A, and the second() overwrites it by a connection B. Once [A's state is updated to ready, Transport wakes up its delegate of WSEngine](https://github.com/daltoniam/Starscream/blob/master/Sources/Transport/TCPTransport.swift#L107) and [WSEngine starts preparing an HTTP upgrade info](https://github.com/daltoniam/Starscream/blob/master/Sources/Engine/WSEngine.swift#L119-L121). However, if [overwriting a connection happens](https://github.com/daltoniam/Starscream/blob/master/Sources/Transport/TCPTransport.swift#L81) before [writing it back to Transport](https://github.com/daltoniam/Starscream/blob/master/Sources/Engine/WSEngine.swift#L122), duplicate upgrade info is sent to connection B in succession. Accordingly, it results in two errors consisting of [an HTTP server error](https://github.com/daltoniam/Starscream/blob/master/Sources/Engine/WSEngine.swift#L168) and [a frame processed error](https://github.com/daltoniam/Starscream/blob/master/Sources/Engine/WSEngine.swift#L179).

By the way, NativeEngine doesn't suffer from this issue. It simply creates two connections.

### Implementation Details 🚧

To prevent this sort of error, this PR introduces a new property called "isConnecting" and forces WSEngine to start connecting only when the property is false.

I believe this change is the most concise solution under the current base architecture, but let me know how you think.

### DIFF
--- a/Sources/Engine/WSEngine.swift
+++ b/Sources/Engine/WSEngine.swift
@@ -24,6 +24,7 @@ FrameCollectorDelegate, HTTPHandlerDelegate {
     private let writeQueue = DispatchQueue(label: "com.vluxe.starscream.writequeue")
     private let mutex = DispatchSemaphore(value: 1)
     private var canSend = false
+    private var isConnecting = false
     
     weak var delegate: EngineDelegate?
     public var respondToPingWithPong: Bool = true
@@ -50,9 +51,10 @@ FrameCollectorDelegate, HTTPHandlerDelegate {
     
     public func start(request: URLRequest) {
         mutex.wait()
+        let isConnecting = self.isConnecting
         let isConnected = canSend
         mutex.signal()
-        if isConnected {
+        if isConnecting || isConnected {
             return
         }
         
@@ -64,6 +66,9 @@ FrameCollectorDelegate, HTTPHandlerDelegate {
         guard let url = request.url else {
             return
         }
+        mutex.wait()
+        self.isConnecting = true
+        mutex.signal()
         transport.connect(url: url, timeout: request.timeoutInterval, certificatePinning: certPinner)
     }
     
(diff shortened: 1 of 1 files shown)


hunble:
i am facing a problem with this, so if we turn off internet and make the first connect() call, it would set "isConnecting" after the internet comes back it does not automatically reconnects and subsequent connect() does not proceed as well because "isConnecting" is true.

yoheimuta:
I created a PR to fix the above problem: https://github.com/daltoniam/Starscream/pull/1053.

PR #937, opened 2021-12-25
Fix compression when headers don't include window bits

zacwest:
### Goals ⚽
When connecting to an aiohttp WebSocket, it excludes the `client_max_window_bits` and `server_max_window_bits` values [when at the default of 15](https://github.com/aio-libs/aiohttp/blob/eda8114ae1246aab135cef5201581025b4fb6fd0/aiohttp/http_websocket.py#L232).

### Implementation Details 🚧
This changes the compressor/decompressor to be created with the default unless otherwise specified. This also fixes a crash when trying to decode the PONG dataframe which has no content.


acmacalister:
Thank you

PR #942, opened 2022-01-26
Fix for start websocket with cookies in initial request

newloran2:
### Goals ⚽
The possibility to configure cookies per connection and not global

### Implementation Details 🚧
in some cases it may be necessary to pass some specific cookie to connection, but all cookies in the initiator request will be replaced by cookies in the cookie storage.
With this verification if the client pass a set of cookies in the initiator request these will used instead the ones in the cookie Storage


BlackWolf:
@daltoniam Just wanted to ask if something prevents this from being merged? It is a major issue for us since it basically means that all custom cookies in the initiator request are overwritten as soon as something in the shared cookie storage matches the URL.

PR #981, opened 2023-09-19
Update Starscream.podspec

pabiagioli:
- Increased the podspec version number to a number that hasn't been used


Starscream.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = "Starscream"
-  s.version      = "4.0.4"
+  s.version      = "4.0.7"
```
bgoncal, line 3: Now it can already be bumped to 4.0.8

PR #1011, opened 2024-03-25
Avoid creating multiple URLSessions

bgoncal:
Apple guidelines recommend that we don't create several URLSession, also, some users may face `POSIXErrorCode(rawValue: 28): No space left on device` when a certain number of URLSessions were already created. I don't know all edge cases that this lib handles, so this is just an initial suggestion

Related: https://stackoverflow.com/questions/67318867/error-domain-nsposixerrordomain-code-28-no-space-left-on-device-userinfo-kcf


bgoncal:
@daltoniam Do you think we can merge this one?

algrid:
It would be great to merge this fix, I also have this problem with too many URLSessions.

PR #1014, opened 2024-04-09
Propagate errors triggered on 'waiting'

pbartolome:
### Goals ⚽
When Starscream uses the `NWConnection` from the Network framework (iOS 12+), some scenarios just fail silently without any type of feedback or callback back to the caller. 

This is mainly due the fact that those errors are being returned when the connection state is `waiting`. According to Apple's documentation

> Connections that are waiting will indicate the reason that the connection couldn't be established in the associated error. These errors are not fatal.

But this state is being reached on 'fatal errors' like connection rejected because SSL pinning failed or timeouts. In those cases, as the `waiting` state is reached, Starscream just ignores the returned error and the client will just wait as it won't receive any callback.

### Implementation Details 🚧

To maintain compatibility with older implementations and to not modify existing behaviour a new `WebSocketEvent` has been added `waiting(Error)`. This will contain the error propagated from the `TCPTransport`


pbartolome:
@acmacalister @daltoniam please take a look and let me known if there are changes needed to merge this PR so we can propagate and handle those errors.

Issue #848, opened 2020-11-12
Compression on control frames.


jmpg93:
I've recently enabled the compression feature, but I've noticed that the default compression handler class (`WSCompression`), does compress every message sent through the socket. Including both data and control messages. 
However, according to [RFC7692](https://tools.ietf.org/html/rfc7692#section-6): _PMCEs operate only on data messages_.

* Pong sent through the socket when **no using** `WSCompression`:
![no_compression_pong_working_as_expected](https://user-images.githubusercontent.com/5618731/98937816-434b8480-24e7-11eb-94b8-bbf4a0f8707e.png)

* Pong sent through the socket when **using** `WSCompression`:
![compression_pong_not_working_as_expected](https://user-images.githubusercontent.com/5618731/98937811-4181c100-24e7-11eb-9486-1cd3f61eb711.png)

In the second case, the socket is closed by the peer, as the socket is automatically sending (what it seems to be) the compressed pong (`6200 00`).

Any ideas? Should the `CompresionHandler` protocol receive control messages? 

Maybe I'm missing something.

PasqualePuzio:
Look at this pull request https://github.com/daltoniam/Starscream/pull/1042 
They came up with a nice solution

PR #1042, opened 2024-10-12
Disable compression on control frames

jonstaff:
As noted in #848, compression should only be applied to data frames. When compression is applied to control frames, specifically to pong messages, a correctly implemented server will fail the websocket connection:

> An endpoint MUST NOT set the "Per-Message Compressed" bit of control
> frames and non-first fragments of a data message.  An endpoint
> receiving such a frame MUST _Fail the WebSocket Connection_.

From the RFC: https://datatracker.ietf.org/doc/html/rfc7692#section-6.1

We noticed that our app was getting disconnected every 20 seconds (when a ping times out). This resolves that problem.

### Issue Link 🔗

https://github.com/daltoniam/Starscream/issues/848

### DIFF
--- a/Sources/Engine/WSEngine.swift
+++ b/Sources/Engine/WSEngine.swift
@@ -119,14 +119,19 @@ FrameCollectorDelegate, HTTPHandlerDelegate {
             if !canWrite {
                 return
             }
-            
+
             var isCompressed = false
             var sendData = data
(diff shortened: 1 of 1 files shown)


PasqualePuzio:
Had the same issue, fixed it the same way :)

PR #1051, opened 2025-02-27
Revert "Fix connection race when using WSEngine"

yoheimuta:
Reverts daltoniam/Starscream#930

We've also identified [an issue with handling the "isConnecting" flag](https://github.com/daltoniam/Starscream/pull/930#issuecomment-2149276411). Once triggered, it doesn't automatically reconnect. For now, I believe we should revert this change.


yoheimuta:
Sorry, I can't reproduce it locally at the moment, so I’ve closed the PR for now. I'll continue investigating, and if I manage to reproduce the issue, I'll either reopen the PR or submit a fix.

